PR #33, opened 2026-06-25
Analysis-tier: Go pclntab function-name recovery (GolangSymbolAnalyzer)

mahaloz:
Port the name-recovery half of Ghidra's GolangSymbolAnalyzer. A new AnalysisPass (s1_pclntab::GoPclntabPass, id `gopclntab`, default-ON, Go-gated) that, when the binary is Go, locates and parses the embedded pclntab and emits a SymFact{Function} per Go function — so main.main / main.compute / runtime.* render NAMED instead of sub_<addr>. The pclntab survives stripping, so this works even on a `-ldflags=-s -w` binary that has no .symtab.

- Locate: .gopclntab/gopclntab section -> runtime.pclntab symbol ->
  validated byte scan of .rodata/.data.rel.ro/.noptrdata/.data
  (GoPcHeader.getPcHeaderAddress + findPcHeaderAddress). Endianness via the
  LE-then-BE magic probe.
- Parse all four magics: go1.2 0xfffffffb (legacy direct-functab), go1.16 0xfffffffa (absolute entry), go1.18 0xfffffff0 / go1.20 0xfffffff1 (32-bit PC-relative entryoff from textStart). Byte layouts from Ghidra's GoPcHeader/GoModuledata/GoFunctabEntry/GoFuncData, cross-checked against the Go runtime source at go1.18. Inline sentinels + go:buildid names skipped. Defensive: any malformation => empty output, never panic.
- Gate: registered in passes_for only for Compiler::Go (the same .go.buildinfo gate the Go no-return list uses, Increment 15), so non-Go binaries are structurally unaffected. Reuses the EXISTING symbol commit arm (idempotent: a real .symtab name still wins). Settable `--option gopclntab on|off` (stages.toml + KUNA_OPTION_NAMES + architecture flag + engine analysis_pass_enabled + kuna_live_value), default-on. settable_count 36 -> 37; catalog fixture + docs/assertions.md regenerated; count fixtures updated.

Tests (both ran, not skipped): 12 hermetic parser units (no `go` — the merge gate: go1.18/1.16/1.2 decode, LE+BE, malformation/defensiveness) + a real-Go e2e (verify_go_pclntab.rs) that `go build -ldflags=-s -w`s a compute()-calls-main() program at runtime (hermetic, skips if no `go`) and asserts main.main/main.compute/runtime.main are recovered as NAMED on the STRIPPED binary, and that `--option gopclntab off` suppresses it.

Gates: make test 675/675 PARITY OK; make test-stages 159/159 PARITY OK; make rust-test green; kuna catalog --check OK. golang-symbols reclassified ⛔->✅ in docs/analysis-port-log.md (Increment 34). No divergence to the parity oracles (Go-gated, real-ELF-path-only).

Claude-Session: https://claude.ai/code/session_01FvRnfTgYKoZvS2TEvdYpWb


mahaloz:
APPROVE

mahaloz:
Closing as **superseded by #31** ("Analysis-tier: Go pclntab function-name recovery"), which was merged into `main` on 2026-06-25. This branch is a duplicate of that work: its commit `fb1654c` is already upstream by patch-id (`git cherry` reports it as applied), the `s1_pclntab/` files are byte-identical to `main`, and rebasing this branch onto `main` drops the commit and yields an empty diff — nothing left to merge. The `gopclntab` pass + option are live in `main`. (The "CONFLICTING" status was only because the branch was ~66 commits stale.)

PR #36, opened 2026-06-25
[PROPOSAL] angr-abnormal-switch-case-case3-0a21b8: multi-predecessor unrolled-guard jump-table recovery

mahaloz:
# [PROPOSAL] abnormal-switch-case-case3 — multi-predecessor unrolled-guard jump-table recovery

**Scope: LARGE — needs human go/no-go before any implementation worker is spent.**

## The problem

angr's `test_decompiling_abnormal_switch_case_case3` (`memmove` @ `0x18003c330`,
`msvcr120.dll`) recovers a clean `switch (count)` with 17 cases. kuna loses the
switch and emits an indirect *call*:

```c
                    /* WARNING: Treating indirect jump as call */
v7 = (void *)(*(code *)((uint8)*(uint4 *)(a2 * 4 + 0x18003c410) + 0x180000000))(a0,a1);
```

The dispatch is an **image-base-relative jump table**
(`target = 0x180000000 + table[index*4]`, 17-entry RVA table at `0x18003c410`)
whose `BRANCHIND` block has **two predecessors with two different bound guards**
(`cmp $0x10 ; jbe` at function entry; `and $0x7 ; test ; jne` on the
backward-copy path that re-enters the dispatch at `0x18003c3fc`). See
`analysis.md` for the full asm + decompilation evidence.

## Why it is not a one-Action feature (Hard Rule 7)

The recovery does not fail for lack of a small helper — it fails inside the
**core S2 jump-table guard analysis**:

- `JumpBasicModel::analyze_guards` (`s2_lift/jumptable.rs:1896`) bounds the table
  by walking backward through **single-predecessor** blocks for the dominating
  guard. With `size_in() > 1` it short-circuits to `check_unrolled_guard`.
- `check_unrolled_guard` (`s2_lift/jumptable.rs:2080`) is an **empty SEAM stub**:
  `// SEAM(structuring): checkCommonCbranch + findMultiequal + liftVerifyUnroll`.
- No bound ⇒ `recover_model_basic` returns `false`
  (`s2_lift/jumptable.rs:2113`) ⇒ `flow.rs:2719` truncates the indirect jump to
  a call.

Neither gated precedent fits: `kuna_loweredswitch.rs` manufactures a table from
a *comparison cascade* (none here); `kuna_switchmodbound` presupposes the guard
walk already melded a path (it did not). Closing the gap means **porting the
upstream multi-predecessor unrolled-guard machinery** into the
`check_unrolled_guard` SEAM — modifying `JumpBasicModel` core, not a single
gated early-return. This trips Hard Rule 7 on two counts (new S2 recovery
infrastructure; touches switch-recovery core beyond a gated early-return).

## angr reference

angr's `JumpTableResolver` / "abnormal switch case" handling resolves jump
tables whose bound check is split across multiple predecessor guards. The kuna
analog is upstream Ghidra's `JumpBasic::checkUnrolledGuard` →
`BlockBasic::findMultiequal` + `JumpBasic::checkCommonCbranch` +
`liftVerifyUnroll` (`decompiler/cpp/jumptable.cc`), which kuna left as the SEAM
stub at `jumptable.rs:2080`.

## Implementation plan (multi-step)

1. **Port `check_unrolled_guard`** in `s2_lift/jumptable.rs`:
   - `checkCommonCbranch` — when the rootblock has multiple predecessors, find a
     CBRANCH common to the guard structure of each predecessor path.
   - `BlockBasic::findMultiequal` (`s2_lift/blockbasic.rs` or `block.rs`) — pair
     up the MULTIEQUAL inputs across predecessors so the switch variable's value
     range is the union of per-path guards.
   - `liftVerifyUnroll` — verify the lifted (unrolled) guard genuinely bounds the
     index on every path before accepting the bound.
2. **Bound the table** from the unioned guard so `find_normalized` yields
   `jrange.size <= maxtablesize`, then let the existing readonly-LOAD
   image-base-relative table model (`jumptable.rs:2011`) read the 17 RVAs.
3. **Gate** the whole behavior behind a new `option abnormalswitch on|off`
   (default OFF while developing; ablation decides the ship default), so default
   output stays byte-identical when off.
4. **Test harness prerequisite (go/no-go):** kuna's loader is ELF-only and
   cannot load this PE. The end-to-end testcase must use a raw **bytechunk**
   (function bytes `0x18003c330..0x18003c895` at their real addresses, proven to
   reproduce the gap in `analysis.md`). The standard pipeline before/after demo
   (`kuna decompile <pe>`) will NOT run on this target.

## Speed / risk assessment

- **Risk: HIGH.** `analyze_guards` / `find_normalized` are core to *every*
  jump-table recovery in the engine. A faithful port must keep the 675/675
  datatest parity intact (several datatests exercise jump tables). The behavior
  must be option-gated so default output is byte-identical until the ablation
  proves it clean.
- **Speed:** guard analysis runs once per BRANCHIND during S2; the unrolled-guard
  walk is bounded (`maxbranch=2`, `maxpullback=2`). Expected negligible per-table
  cost, but must be measured on a jump-table-heavy corpus before any default-on.
- **Surface:** estimated 2 ported-core files (`jumptable.rs`, a block helper) +
  `options.rs`/`architecture.rs`/`stages.toml` registration anchors, plus the
  bytechunk test. Larger than one new module.

## Proposed option

`option abnormalswitch on|off` — "Recover image-base-relative jump tables whose
bound guard is split across multiple dispatch predecessors (angr's abnormal
switch case)." `source_decompiler=angr`,
`inspiration="test_decompiling_abnormal_switch_case_case3; JumpTableResolver / abnormal switch case; 0x18003c330"`,
`change_kind=structure-recovery`.

## Recommendation

File this draft and pause for human go/no-go. On approval, re-dispatch an
implementation worker on `feat/angr-abnormal-switch-case-case3-0a21b8` to port
`checkUnrolledGuard`/`findMultiequal`/`liftVerifyUnroll` behind
`option abnormalswitch`, with the bytechunk harness for the end-to-end test.

### DIFF
--- a/docs/features/abnormal-switch-case-case3-0a21b8/analysis.md
+++ b/docs/features/abnormal-switch-case-case3-0a21b8/analysis.md
@@ -1,0 +1,111 @@
+# abnormal-switch-case-case3 — analysis
+
+## Opportunity
+
+- angr testcase: `test_decompiling_abnormal_switch_case_case3 :: 0x18003c330`
+- Binary: `binaries/tests/x86_64/windows/msvcr120.dll` (PE32+ / x86-64), function `memmove`
+- angr 9.2.213
+
+## What angr does better
+
+angr recovers a clean, fully-structured `switch (ptr6)` with 17 cases (0..16),
+each case a small tail-copy body returning `addr`:
+
+```c
+switch (ptr6) {
+case 0:  return addr;
+case 1:  *((char *)ptr) = *((char *)p); return addr;
+...
+case 16: *((int128_t *)ptr) = *((int128_t *)p); return addr;
+}
+```
+
+kuna **loses the switch entirely** and renders the indirect dispatch as an
+indirect *call*:
+
+```c
+                    /* WARNING: Treating indirect jump as call */
+v7 = (void *)(*(code *)((uint8)*(uint4 *)(a2 * 4 + 0x18003c410) + 0x180000000))(a0,a1);
+return v7;
+```
+
+(Both decompilations are saved verbatim in `angr-vs-kuna.txt`.)
+
+## The exact construct
+
+`memmove` dispatches the small-count (`count <= 16`) tail through an
+**image-base-relative jump table**:
+
+```
+18003c3f9:  mov  %r11,%r10
+18003c3fc:  lea  -0x3c403(%rip),%r9      # r9 = 0x180000000  (image base, a pure computed constant)
+18003c403:  mov  0x3c410(%r9,%r8,4),%eax  # eax = *(uint32*)(0x18003c410 + r8*4)  ; 17-entry RVA table
+18003c40b:  add  %r9,%rax                 # rax = 0x180000000 + table[r8]
+18003c40e:  jmp  *%rax                     # BRANCHIND  -> the switch
+```
+
+The index `r8` (`count`) is bounded by `cmp $0x10,%r8 ; jbe 0x18003c3f9` at the
+**function entry** — far from the `jmp`. Crucially the dispatch block at
+`0x18003c3fc` is *also* entered from `0x18003c77a` (`jmp 0x18003c3fc`) on the
+backward-copy path, where `r8` is bounded by a **different** guard
+(`and $0x7,%r8 ; test %r8,%r8 ; jne ...`). So the `BRANCHIND` has a
+**multi-predecessor dispatch with no single dominating bound guard**. This is
+precisely what angr names the "abnormal switch case."
+
+## The owning stage and why kuna fails
+
+Stage: **S2 flow/switch model** — jump-table recovery
+(`decompiler/crates/kuna-decomp/src/s2_lift/jumptable.rs`,
+`.../flow.rs`). kuna IS Ghidra's decompiler; the `BRANCHIND` already exists and
+the basic jump-table model is invoked, but it **declines** and the fallback in
+`flow.rs:2719` (`self.data.warning("Treating indirect jump as call", ...)`)
+fires.
+
+Root cause (traced in `jumptable.rs`):
+
+1. `JumpBasicModel::analyze_guards` (`jumptable.rs:1896`) bounds the table by
+   walking **backward from the BRANCHIND's parent through single-predecessor
+   blocks** to find the dominating `cmp/jbe` guard.
+2. Here the dispatch block has **two** predecessors, so the
+   single-predecessor walk short-circuits and calls
+   `check_unrolled_guard` (`jumptable.rs:1927`).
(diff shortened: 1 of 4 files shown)


mahaloz:
## Closing — implemented by #102 (`switchmultipred`), default-on since #110

Re-triaged against today's `main` by running kuna on the actual target and ablating. Both blockers this proposal flagged are gone.

**The loader go/no-go is stale.** The proposal said kuna's ELF-only loader could not open `msvcr120.dll`, so the demo needed a bespoke bytechunk harness. PR #40 (PE linked-exe loading, merged ~30 min after this proposal was filed) removed that. Today:

```
$ kuna decompile .../x86_64/windows/msvcr120.dll 0x18003c330 --addr
real 0m43.5s   exit 0   408 lines   (real memmove body, not a stub)
```

**The gap itself is closed by default.**

|---|---|---|---|
| `switch` | 0 | **1** | 1 |
| `case` labels | 0 | **17** (`case 0` … `case 0x10`) | 17 |
| `Treating indirect jump as call` | 1 | **0** | 0 |

Ablation is causal — `--option switchmultipred off` reproduces this proposal's bug verbatim:

```
$ kuna decompile msvcr120.dll 0x18003c330 --addr --option switchmultipred off
367 lines, 0 switch, 0 cases, 1 warning:
    /* WARNING: Treating indirect jump as call */
    return (void *)(*(code *)((uint8)v4 + 0x180000000))(a0,a1);
```

Case bodies 6/7/8/9/0x10 were spot-diffed against angr's — structurally identical tail copies (case 8 = one 8-byte copy, case 0x10 = 16-byte copy), so the recovered table is the right table, not an over-bound one.

**Why #102 is this proposal's implementation under a different name.** #102's body opens *"Implements proposal #36 (`abnormal-switch-case-case3-0a21b8`)"*, names `0x18003c330` / `0x18003c410`, and the shipped option carries `inspiration = "test_decompiling_abnormal_switch_case_case3; JumpTableResolver / abnormal switch case; 0x18003c330"`. The `check_unrolled_guard` SEAM stub this proposal wanted filled is now a real body at `decompiler/crates/kuna-decomp/src/p2_lift/jumptable.rs:2437` (call site :2071), with stage test `tests/stages/ghangr-switchmultipred-memmove.xml` and 7 `switchmultipred` keys in `docs/baseline-stages.json`. #110 flipped it default-on — *stronger* than this proposal asked for (it proposed `abnormalswitch` default-off, "ablation decides the ship default").

**Residual (out of scope here):** kuna emits 1 goto (`label_18003c6c0`, in the large-copy path outside the switch) where angr emits 0, and 408 loc vs angr's 467. Neither is part of this proposal's stated criterion — this was a jump-table-recovery proposal and that criterion is fully met.

PR #37, opened 2026-06-25
ptx-fix-output-parameters: ET_REL relocatable-object loader (option relocobjects)

mahaloz:
# `relocobjects`: load ELF relocatable objects (`ET_REL` / `.o`)

Closes the angr-vs-kuna gap `test_decompiling_ptx_fix_output_parameters::fix_output_parameters`
(`ptx.o`, angr 9.2.213). Analysis: [`analysis.md`](analysis.md) · approved proposal:
[`proposal.md`](proposal.md).

## The gap

`ptx.o` is a **relocatable ELF object** (`ET_REL`) with no `PT_LOAD` program headers. kuna's
faithful `LoadImageBfd` port (`kuna-analysis/loadimage_object.rs`) builds its byte map **only** from
`PT_LOAD` segments, so a `.o` mapped **zero bytes** and *every* function failed to lift
(`Unable to load N bytes at ...`) — kuna emitted **no output at all**. angr's CLE ELF relocatable
backend decompiles it fully with resolved names.

## Mechanism

A new kuna file `kuna-analysis/src/s1_loader/elf_reloc.rs`. For an `ET_REL` object
(`kind()==Relocatable` + empty `segments()`), `ObjectLoadImage::from_bytes` takes a new
`from_relocatable` path that:

1. **lays out** each `SHF_ALLOC` section at a non-overlapping VMA above `0x400000` (angr's CLE
   default — `fix_output_parameters` lands at `0x400660`, matching angr), respecting alignment;
   `.bss`/`NOBITS` is zero-filled;
2. **applies** the `.rela.*` relocations into the snapshotted bytes — `R_X86_64_PC32`/`PLT32`
   (`S + A − P`), `R_X86_64_64`/`32`/`32S` (`S + A`); an unhandled kind degrades with a logged
   warning rather than miscompiling;
3. **rebases** defined symbols to their load VMA and **binds undefined externs** (PLT-relative call
   targets are named even when gcc emits them `STT_NOTYPE`) to synthetic addresses, so calls render
   by name (`strlen(...)`, `__sprintf_chk(...)`, `__ctype_b_loc(...)`, …).

The linked `ET_EXEC`/`ET_DYN` `PT_LOAD` path is untouched (byte-identical).

## Option

`relocobjects` (default **on** — it only affects `.o` files, which the loader otherwise cannot load
at all). The loader runs at `load file`, **upstream** of the per-function option machinery, so the
[...]

PR #39, opened 2026-06-25
[PROPOSAL] angr-tr-build-spec-list-ea32be: SAILR condition folding / goto reduction (S7, large)

mahaloz:
# [PROPOSAL] SAILR-style condition folding / crossing-edge goto reduction (S7)

**Opportunity:** `test_decompiling_tr_build_spec_list::build_spec_list`
**Binary:** `binaries/tests/x86_64/decompiler/tr.o` (coreutils `tr`, x86-64; ET_REL)
**Proposed option:** `condfold` (S7 region structuring, default-OFF)
**Scope:** **LARGE** — needs human go/no-go before any implementation worker is spent.

## The problem

For `build_spec_list`, kuna's full-pipeline output (`kuna-fullpipeline.c`) is structurally
close to angr's (`angr-vs-kuna.txt`) — DWARF types and loops match, `__stack_chk_fail` is
already no-return — **except** that kuna emits **11 gotos** with *crossing edges* where angr
emits **4**, because angr collapses short-circuit control flow into compound conditions.

Concretely, the bracket-class dispatch:

```c
// angr — one compound condition, no crossing goto
if (!(char)es_match(ptr, cur, '[') ... )
if (!(char)es_match() && !(char)es_match() || !(char)find_closing_delim(ptr, v11, ..., &v7)) {
LABEL_400d08: ...
}
```

```c
// kuna — nested if/else with a crossing-edge goto back into the sibling branch
v2 = es_match(v15, v10, ':');
if (v2) { ... find_closing_delim ... if (!v2) goto label_4024b8; ... }
else {
  v2 = es_match(v15, v14, '=');
  if (v2) goto label_40242e;     // <-- jumps from the else branch back into the then branch
label_4024b8: ...
}
```

The `goto label_40242e` is a crossing edge that angr removes by recovering
`es_match(':') || es_match('=')` as a single compound condition.

## The angr reference

angr's decompiler (Basque et al., *SAILR*, USENIX Security 2024) produces this via:

- **`RegionIdentifier` + `Phoenix` structuring** — synthesizes compound `&&`/`||` condition
  nodes when two condition blocks share a successor (the short-circuit schema).
  (`angr/analyses/decompiler/structuring/phoenix.py`, `_make_condition_node` /
  `_match_acyclic_schemas`.)
- **SAILR optimization passes** — `CrossJumpReverter`, `DuplicationReverter`,
  `ConditionConstantPropagation` (`angr/analyses/decompiler/optimization_passes/`) that undo
  compiler control-flow duplication so the crossing edges disappear.

Ghidra (and thus kuna today) structures with the BlockGraph collapse rules
(`blockaction.cc`), which do **not** synthesize short-circuit compound conditions across
distinct basic blocks — they emit a goto for the crossing edge instead.

## Why this is LARGE (Hard Rule 7)

- It is **S7 structuring** logic, not a single gated early-return: it must add condition
  synthesis to the region collapse and/or a pre-structuring deoptimization pass.
- It is a **new pass type / infrastructure**, not one `Action`/`Rule` modeled on
  `kuna_loweredswitch.rs` (which manufactures an S2 artifact and reuses the existing
  structurer). Compound-condition recovery has no existing seam to reuse.
- It will likely touch **>3 ported-core anchor files** (`blockaction.cc`/the S7 region
  collapse, `block.cc` for a new compound-condition block kind, the C-printer for `&&`/`||`
  emission, plus the option/registration anchors) and **>1 module**.

Any one of these alone triggers the LARGE classification.

## Proposed multi-step implementation plan (for the approved worker)

1. **Detect the short-circuit schema** in the S7 region graph: two `BlockCondition`s whose
   true/false edges share a target (the `A && B` / `A || B` shapes), pre-collapse — port of
   `Phoenix._make_condition_node`. Gate behind `option condfold`.
2. **Synthesize a compound-condition region node** (a new `BlockCondition` whose predicate is
   `BoolAnd`/`BoolOr` of the two operands) and rewire edges so the crossing goto becomes a
   structured edge.
3. **C-printer support** for emitting the folded `&&`/`||` (likely already present for
   in-block boolean ops; verify it renders for the synthesized node).
4. **Crossing-edge deoptimization** (SAILR `CrossJumpReverter`) as a follow-up sub-step if
   step 1–2 do not remove all crossing edges on the corpus.
5. **Stage test** `tests/stages/ghangr-tr-build-spec-list-ea32be.xml`: pass 1 (`option
   condfold off`) asserts the crossing `goto`; pass 2 (default-on for the test) asserts the
   compound condition and goto reduction.

## Speed / risk assessment

- **Risk: HIGH.** S7 structuring changes are the riskiest in the engine — the 675-datatest
  parity gate is very sensitive to block-collapse order. Expect a large ablation; the option
  must ship **default-OFF** opt-in regardless of speed.
- **Speed:** an extra schema-match scan over region nodes is O(blocks); modest, but must be
  measured on `build_spec_list` off-vs-on per Hard Rule 6 before shipping.
- **Generalizes:** yes — short-circuit `&&`/`||` lowering is ubiquitous in `-O2` GCC/Clang
  output, so a correct port helps far beyond this one function.

## Recommendation

Approve as a **dedicated S7-structuring implementation worker** (not this single-feature
slot). The smallest viable first cut is steps 1–3 (compound-condition synthesis for the
acyclic short-circuit schema), gated `option condfold`, default-off, with the stage test
above; defer the SAILR deoptimizers (step 4) to a follow-up if needed.

> Note: the *target binary itself* (`tr.o`) cannot be loaded by kuna (ET_REL loader gap,
> proposal PR #37). The approved worker should use the linked standalone ELF approach
> documented in `analysis.md` (`stubs.c` + `tr_synth.elf`) or a relocated bytechunk
> (`probe.xml`) for its stage test, exactly as done here.


mahaloz:
## Re-triage against today's `main` — still a real gap; implementation dispatched

Keeping this open. Re-measured, and the headline number moved, but **not for the reason it looks like**, and the named defect is still fully present.

### The target loads now

`tr.o` loads directly today (`relocobjects`, #37, default-on) — it did not at proposal time. `build_spec_list` @ `0x400bc0`, matching angr's address.

### On `tr.o` directly

| | proposal baseline | today, defaults | angr 9.2.213 |
|---|---|---|---|
| gotos | 11 | **5** | 4 |
| loc | 122 | 116 | 128 |
| folded `\|\|` | 0 | 1 | (compound) |

### But the 11 → 5 comparison is cross-binary and cross-fidelity

The baseline `11` was measured on the **synthetic linked ELF** (`stubs.c` + `tr_synth.elf`); today's `5` is on **ET_REL `tr.o`**, whose output is materially lossier — signature degrades from `build_spec_list(E_string *a0, Spec_list *a1)` to `(int8 *a0, unsigned long a1)`, string literals degrade to raw addresses (`v5 = 0x404858;` instead of `"missing character class name '[::]'"`), and calls lose arguments (`es_match()` with none, `find_closing_delim(a0,v17)` with 2 of 4). Fewer gotos partly reflects less recovered structure.

### Apples-to-apples, the proposal's defect is 100% present

Rebuilding this branch's own measurement vehicle bit-for-bit (`gcc -no-pie -nostdlib -nostartfiles -static -Wl,-Ttext-segment=0x400000 tr.o stubs.c`; `build_spec_list` lands at `0x402370`, exactly as `analysis.md` documents), today's defaults emit:

```c
if (es_match(es,v12,'=')) goto label_40242e;   // THE crossing edge, else arm -> then arm
```

with **zero folds** (0 `||`, 0 `&&`).

### Correction to an earlier read: the existing fold is not landed P8 work

With **all 11 P8 goto-quality options forced off**, kuna still folds the short-circuit condition — it just renders it in negated polarity as `&&`:

```c
else if ((es_match() == '\0') && (es_match() == '\0'))
```

So the fold on `tr.o` comes from the pre-existing ported-Ghidra `CollapseStructure::ruleBlockOr`, which kuna has had since the C++ port — not from #69/#79/#85/#103/#110. `branchflip` only flips polarity (`&&` → `||`); it does not create the fold. **Measured worth of the entire landed P8 family on this function: exactly 1 goto** (tr.o 6→5, synth 7→6).

### Root cause, now localized

The else block carries `result_idx = &v10;` *before* the condition, so the `isComplex` gate declines the fold. angr's Phoenix folds it anyway. That is precisely this proposal's thesis — Ghidra-style collapse rules can't fold across a non-trivial sibling block — and it is generalizable, not noise.

### What changed since 2026-06-25: this is no longer LARGE

This PR was ruled LARGE on the premise that compound-condition recovery *"has no existing seam to reuse"*. That premise is now obsolete. `p8_structure/region_structurer.rs` (default path, `regionstructure` on since #85/DIV-12) already contains `match_acyclic_short_circuit_conditions()` (:865) — the port of angr `phoenix._match_acyclic_short_circuit_conditions` — and `try_block_or()` (:886). The remaining work is relaxing the complexity gate to tolerate a bounded side-effect-free prefix, i.e. an **increment on an existing schema**, not a new pass type.

### Cross-reference

Now-closed #53 (`true_ubuntu1804::usage`) has a measured residual of exactly **+2 ifs** vs angr from this same root — angr short-circuits `if (!v30 || !strncmp(v30, "en_", 3))` where kuna emits two nested ifs, twice. That is a second witness for `condfold`.

An implementation worker is running on branch `feat/condfold` (default-off option, two-pass stage test, four gates, speed + decbench measurement).

mahaloz:
## Implementation is up: #196

This proposal and its sibling turned out to share **one** root cause — Ghidra's `BlockBasic::isComplex` >2-statement readability budget gating the short-circuit fold in `CollapseStructure::rule_block_or` (and its verbatim twin `try_block_or` in the region structurer). Proposal #39 (`tr::build_spec_list` crossing edge) and proposal #56 (`newbury::main` SUID-guard goto) are two symptoms of that single gate.

They were implemented separately as #193 and #194, which collided (same struct field, same builder method, same gate line, same ElementId). Both are now closed and consolidated into **#196** — one option `condfold off|on|wide`, default-off, ElementId 4113, whose admission predicate is the union of the two rules.

Both criteria are met and measured against `main` @ `d09f21ce`:

- **#56** — `newbury::main` 1 goto / 1 label → **0 / 0** at `on`, matching angr.
- **#39** — the crossing `goto label_40242e` is gone at `on`; angr's exact goto count (4) lands at `wide`.

Notably, both proposals' **LARGE / "new pass type" verdicts were wrong**, and for the same reason: they were filed on 2026-06-25, before `regionstructure` (#85, DIV-12) made the Phoenix/SAILR region structurer the default emit-driving path. Proposal #56's step 1 — *"promote `kuna_regionid.rs` from a read-only console surface into an emit-driving structurer"* — was already done. Neither needed a new pass type; each needed one gate relaxation. Nor was a de Morgan rewrite or a purity analysis required: `negate_condition_rec` already distributes the NOT and **never swaps operands**, so #56's worry about reordering the `getuid`/`geteuid`/`getegid`/`getgid` predicates does not arise.

Leaving this open until #196 merges.

PR #43, opened 2026-06-25
[PROPOSAL] angr-od-else-simplification-5194c9: earlyreturn — early-return guard hoisting (else simplification)

mahaloz:
# [PROPOSAL] `earlyreturn` — angr-style early-return guard hoisting ("else simplification")

**Opportunity:** angr `test_od_else_simplification` · function `skip` · `od_gccO2.o` (x86_64)
**Proposed option:** `earlyreturn` (default-off, opt-in) · **Scope: LARGE** (human go/no-go)

## The problem

For a function whose first action is an argument-validity check, angr emits an idiomatic
early-return guard and de-indents the body; kuna wraps the entire body in an inverted
condition. See `analysis.md` and the captured side-by-side in `angr-vs-kuna.txt`.

```c
// angr (desired)                         // kuna (today)
if (!cur)                                 if (a0 != 0) {
    return 1;                                 ...whole body...
v3 = 0;                                    }
... body de-indented ...                   v14 = 1;
                                           label: if (canary) __stack_chk_fail();
                                           return v14;     // !a0 path lands here
```

The angr test (`cstyle_ifs=False`) asserts the **first** `if` is `if (!a0) { return 1; }`
with no else.

## angr reference passes

1. `optimization_passes/return_duplicator_{base,high,low}.py` — duplicate the shared
   epilogue (`__stack_chk_fail` guard + `return 1`) into the `!a0` path so it becomes a
   standalone *terminating* return.
2. `region_simplifiers/ifelse.py` `IfElseFlattener` — when the if's true-branch always
   terminates and the false-branch does not, drop the `else` and re-parent the body as a
   sibling after the `if`.

Both run on angr's mutable region tree.

## Why this is LARGE (not one option-gated Action/Rule)

- kuna has **no angr-style mutable region tree** at emit time. The if/else shape is fixed by
  Ghidra's block-graph collapse Structurer in
  `decompiler/crates/kuna-decomp/src/s8_structure/blockaction.rs` (~3600 lines, S7/S8 region
  recovery).
- Reproducing angr's output requires **two cooperating transformations** woven into that
  Structurer: (a) return/epilogue duplication into the guard path, and (b) else removal +
  body de-indent — i.e. a **new structuring pass type**, not a single `Action`/`Rule` like
  `kuna_loweredswitch.rs`.
- It touches S7/S8 structuring/region code well beyond a single gated early-return → Hard
  Rule 7 LARGE. A decider subagent independently confirmed `scope: large`.

## Proposed multi-step implementation plan (for the approved implementation worker)

1. **Pattern detection (S8).** In/after `CollapseStructure`, identify a top-level
   `BlockIf { cond, body }` whose *false* (fall-through) path reaches the function's shared
   return epilogue with no side effects other than constant return-value setup.
2. **Epilogue duplication.** Synthesize a duplicated terminating return (carrying the
   constant from the false path, e.g. `return 1`) into a new guard block `if (!cond) return
   K;` — the analogue of angr's return-duplicator.
3. **Restructure.** Replace the body-wrapping `BlockIf` with the guard followed by the
   de-indented body as siblings (analogue of `IfElseFlattener`). Preserve canary handling so
   `make test` / `make test-stages` parity is unaffected when the flag is off.
4. **Gate.** New `earlyreturn` bool on the architecture struct (default off in reset),
   registered action early-returns when off so default output is byte-identical;
   `settableTable` row + `options.rs` registration; LLM-discoverable.
5. **Verify.** Two-pass `tests/stages/ghangr-od-else-simplification-5194c9.xml` (off = body-wrapped
   bug, on = early-return guard). Ablation over the 675 datatests; default decision gated by
   the speed budget.

## Speed / risk assessment

- **Risk:** structuring rewrites are the highest-blast-radius part of the pipeline; an
  over-broad pattern could change many functions. Mitigate by a tight pattern (top-level
  single-guard + constant-return shared epilogue only) and keeping it **default-off opt-in**
  initially, with a full 675-assertion ablation before any DIV flip.
- **Speed:** one extra structuring walk per function; expected within the +5% budget but
  must be measured on the target (`skip`) before any default-on decision.
- **Loader caveat:** the opportunity binary is ET_REL (loader gap, PR #37); the
  implementation worker should validate via the link-bypass ELF or a bytechunk stage test.

## Relation to existing proposals

Distinct from `condfold` (#39: short-circuit `&&`/`||` folding + crossing-edge goto
reduction) and `elf-reloc-objects` (#37: the ET_REL loader). Same broad SAILR S7/S8
region-structuring family, but a separate, specific transformation (early-return guard
hoisting / else flattening). Not covered by any existing option (`returnpair`,
`sparcstructret`, `stackguard` are unrelated).

### DIFF
--- a/docs/features/od-else-simplification-5194c9/analysis.md
+++ b/docs/features/od-else-simplification-5194c9/analysis.md
@@ -1,0 +1,71 @@
+# Analysis — angr `test_od_else_simplification` (`skip` in `od_gccO2.o`)
+
+## The gap
+
+angr's decompiler emits the early-exit guard of `skip` as the **first** statement and
+de-indents the body:
+
+```c
+if (!cur)
+    return 1;
+v3 = 0;
+v4 = 1;
+while (in_stream) { ... }       // body de-indented, no else wrapping
+```
+
+kuna instead wraps the **entire** function body inside `if (a0 != 0) { ... }` and lets the
+`!a0` path fall implicitly through to a shared epilogue:
+
+```c
+if (a0 != 0) {
+    ...whole body...
+}
+v14 = 1;
+label_403d25:
+if (canary != ...) __stack_chk_fail();
+return v14;                      // the !a0 path lands here -> return 1
+```
+
+The angr test (`cstyle_ifs=False`) asserts the first `if` is `if (!a0) { return 1; }` with
+**no else** — i.e. an early-return guard, not a body-wrapping condition.
+
+## Why angr is better here
+
+The early-return guard is the idiomatic shape for a function whose first action is an
+argument-validity check. kuna's body-wrapping `if (a0 != 0) { ... }` pushes the entire
+function one indent level deeper and obscures the guard.
+
+## angr's mechanism (read from angr source)
+
+Closing this gap is **two cooperating angr passes**, both operating on angr's mutable
+region tree (`ConditionNode` / `SequenceNode` / `LoopNode`):
+
+1. **Return duplication** — `optimization_passes/return_duplicator_*.py`. The shared
+   canary+`return 1` epilogue is duplicated into the `!a0` guard path so that path becomes a
+   standalone, *terminating* `return 1`.
+2. **If/else flattening** — `region_simplifiers/ifelse.py` `IfElseFlattener`. Once the if's
+   true-branch always terminates (returns) and the false-branch does not, it removes the
+   `else` node and re-parents the body as a sibling **after** the `if`, de-indenting it.
+
+## Owning stage in kuna
+
+kuna has **no angr-style mutable region tree at emit time**. The if/else shape is decided by
+Ghidra's **block-graph collapse Structurer** — `CollapseStructure` / the block-if-else rules
+in `decompiler/crates/kuna-decomp/src/s8_structure/blockaction.rs` (~3600 lines), Ghidra's
+S7/S8 region recovery. Reproducing angr's output requires re-creating both halves of angr's
+mechanism (return duplication into the guard path + else removal + body de-indent) inside
+that Structurer, plus a likely emit-side change.
+
+## Verdict
+
+**LARGE** (Hard Rule 7): a new structuring pass *type* that touches S7/S8 structuring code
+well beyond a single gated early-return. Not modelable as one Action/Rule like
+`kuna_loweredswitch.rs` (an S2 detection pass). A decider subagent independently confirmed
+`scope: large` (see `record.json` → `decisions`). This is the same S7/S8 SAILR
+region-structuring class consistently parked LARGE in the five prior ET_REL opportunities
+(`condfold`, `elf-reloc-objects`, …).
+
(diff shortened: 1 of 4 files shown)


mahaloz:
## Closing — implemented by #124/#137 (`returndup`) + #96 (`ifelseflatten`); #147 developed the sibling case

Re-triaged against today's `main` by running kuna on `od_gccO2.o::skip` and sweeping options.

**The ET_REL caveat is stale.** `od_gccO2.o` loads directly today (`relocobjects`, #37, default-on) — 130 functions, `skip` @ `0x401fc0`, matching `record.json`. No synth-ELF link bypass needed.

**The criterion is met, opt-in.** This proposal's criterion is purely structural and stated twice: *"the angr test (`cstyle_ifs=False`) asserts the **first** `if` is `if (!a0) { return 1; }` with no else"*, body de-indented. With `--option returndup on` (also reachable via `--mode aggressive`) kuna emits exactly that — line 23 `if (a0 == 0) { … return 1; }`, no else, whole body de-indented one level, ending `return v11;`.

On stock defaults the first `if` is still the complained-of `if (a0 != 0) { …whole body… } … return 1;`.

**Default-off is precisely what this proposal asked for.** Header: *"Proposed option: `earlyreturn` (default-off, opt-in)"*; plan step 4: *"default off in reset … so default output is byte-identical"*; risk section: *"keeping it default-off opt-in initially, with a full 675-assertion ablation before any DIV flip."* `returndup` being default-off is a measured decision (DIV-18 revert: −976 aggregate GED-perfect on the unselective version; #137 then added the const-return gate), not an oversight.

**The covering options are the two angr passes this proposal named by filename.** It listed (1) `optimization_passes/return_duplicator_{base,high,low}.py` and (2) `region_simplifiers/ifelse.py::IfElseFlattener`, and said kuna needs both halves:

- #124 — *"returndup — duplicate shared bare-epilogue returns for early-return recovery (**angr ReturnDuplicatorHigh**)"*, refined by #137.
- #96 — `kuna_ifelseflatten.rs`, the `IfElseFlattener` port.

Ablation confirms `returndup` alone is the actor.

**One correction for the record.** #147 ("s8: earlyreturn …") is titled *"develops #43"* and its `docs/options.md` symptom rows quote this proposal verbatim ("leading argument-validity check wraps the whole body instead of an early return", "inverted diamond"). But measured on *this* target, `earlyreturn` is a **no-op**: `--option earlyreturn off` is byte-identical to default on `skip`, and across the whole binary it changes 1/130 functions (`decode_one_format`), not `skip`. Its per-edge const-phi gate doesn't match `skip`'s return shape. So #147 generalized the idea and won elsewhere (+47 perfect / −576 GED, DIV-23); the option that closes *this* function is `returndup`.

**Residual (cosmetic):** with `returndup on` the hoisted guard block carries a dead `dat_402748 = dat_402748;` self-assignment, so the guard isn't literally the two-token `if (!a0) return 1;` angr prints. That self-assignment is pre-existing on defaults too — a separate merge/copy rendering issue, not part of this proposal.

PR #47, opened 2026-06-25
[PROPOSAL] angr-mv0-main-8def54: structure multi-latch loops (keep switch inside the getopt loop)

mahaloz:
# [PROPOSAL] angr-mv0-main-8def54 — structure multi-latch loops (keep the switch inside the getopt loop)

**Status:** draft / awaiting human go-no-go. **Do not implement until approved.**

## The problem

On coreutils `mv` `main` (`/home/mahaloz/github/angr-dev/binaries/tests/x86_64/mv_0`,
fn `main` @ `0x402c40`), kuna fails to recover the `getopt_long` option-parsing
loop as a structured `while`. The loop has one header and **many latches** (each
of the 11 `switch` cases back-edges to the `getopt_long` block). kuna's S8
structurer degrades it to a goto-loop: it labels the header (`label_2d00:`),
**hoists the whole switch out of the loop** to the function tail, and synthesizes
`goto label_2d00;` from every case.

Result vs angr (which keeps the switch inside `while(true)`):

| metric | angr | kuna |
|---|---|---|
| gotos | 10 | 29 |
| labels | 9 | 16 |
| loops | 3 | 2 |

See `analysis.md` and `angr-vs-kuna.txt` for the full side-by-side.

## Why this is LARGE (not a one-Action feature)

The gap is in the **S8 collapse fixed point**
(`decompiler/crates/kuna-decomp/src/s8_structure/blockaction.rs`):
`CollapseStructure` / `TraceDAG` / `LoopBody` / `ruleBlockWhileDo` /
`emitLikelyEdges`. `ruleBlockWhileDo` collapses a loop only with a single
structured back-edge; multi-latch loops get their extra back-edges marked
unstructured and the loop falls back to `BlockGoto`, which in turn prevents the
`BlockSwitch` from nesting under the loop. A no-op gated Action cannot force the
existing collapse to keep the switch inside — the multi-latch normalization has to
happen **inside** the structuring fixed point. This touches S7/S8 region/structuring
code well beyond a single gated early-return and likely > 3 ported-core anchor
files. A decider subagent (Opus 4.8, high confidence) classified it LARGE.

## angr reference

- angr `RegionIdentifier` — already ported as
  `decompiler/crates/kuna-decomp/src/s7_regions/kuna_regionid.rs` +
  `kuna_regiongraph.rs` (S7).
- angr Phoenix/Dream-style `Structurer` / `RegionSimplifier` loop refinement:
  `_refine_loop`, break/continue insertion, single-latch normalization — the passes
  that let angr present one back-edge per loop body and render the rest as
  `continue`.

1. **Detect** loop headers with N>1 in-loop back-edges (latches) in the S8 block
   graph, gated behind a new option (default OFF while developing).
2. **Latch unification:** route the N back-edges through a synthesized common
   latch / pre-header block so the loop presents a single structured back-edge to
   `ruleBlockWhileDo`; the per-case jumps become `continue`.
3. **Switch nesting:** teach the `BlockSwitch` path to allow case bodies whose tail
   is a loop back-edge to nest under the enclosing `BlockWhileDo` instead of being
   hoisted/gotos'd.
4. **(Alternative / larger)** route this loop shape through the already-ported S7
   `kuna_regionid` region structurer and bridge its output back into the S8 block
   model — reuses angr's loop refinement directly, but is a bigger integration.
5. Gate behind the option, add a `tests/stages/ghangr-mv0-main-8def54.xml` two-pass
   testcase (off = goto-hoist bug, default/on = structured while), regenerate the
   stage baseline, and measure decompile speed (standing requirement).

## Speed / risk assessment

- **Speed: medium risk.** Latch unification runs inside the S8 collapse fixed point
  on every function with a loop; a naive per-loop-header back-edge scan/merge can
  re-trigger the fixed point and add passes over the block graph. Must benchmark on
  the target (budget default +5%) and likely ship **default-OFF opt-in** until the
  ablation + speed gate are clean.
- **Correctness: medium-high risk.** Changing loop collapse / switch nesting can
  alter structuring on many functions. Requires the full 675-assertion ablation to
  stay PARITY OK; almost certainly default-OFF at first.
- **Anchor files:** `blockaction.rs` (collapse + switch nesting), the block-graph
  model, and option/architecture/stages.toml registration — > 3 ported-core files.

## Proposed option

- Option name: **`multilatchloop`**  (clean; decider's suggestion)
- Next free ElementId: **4021** (highest currently used is 4020).
- `change_kind`: `structure-recovery`
- `source_decompiler`: `angr`;
  `inspiration`: `test_decompiling_mv0_main; RegionIdentifier/Structurer loop refinement (single-latch normalization); main`

### DIFF
--- a/docs/features/mv0-main-8def54/analysis.md
+++ b/docs/features/mv0-main-8def54/analysis.md
@@ -1,0 +1,86 @@
+# Analysis — `test_decompiling_mv0_main::main` (coreutils `mv`, x86_64)
+
+Binary: `/home/mahaloz/github/angr-dev/binaries/tests/x86_64/mv_0`
+Function: `main` @ `0x402c40`  (angr 9.2.213)
+
+## The gap (what angr does better)
+
+`main` is GNU coreutils `mv`'s option-parsing entry: a `getopt_long` loop with an
+11-case `switch` on the returned option character.
+
+| metric | angr (ref) | kuna | delta |
+|---|---|---|---|
+| loc | 284 | 243 | |
+| **gotos** | **10** | **29** | kuna ~3× more |
+| **labels** | **9** | **16** | kuna ~2× more |
+| **loops** | **3** | **2** | kuna recovers one fewer |
+| switches | 1 | 1 | |
+| cases | 11 | 11 | |
+
+Signals emitted by the compare tool:
+- ref has fewer gotos (10 vs 29)
+- ref has fewer labels (9 vs 16)
+- ref recovered more loops (3 vs 2)
+- kuna emitted a recovery-failure marker
+
+### The one concrete structural difference
+
+angr renders the getopt loop natively, with the `switch` **inside** the loop:
+
+```c
+while (true) {
+    v17 = getopt_long(a0, a1, "bfint:uvS:TZ", &long_options.name, NULL);
+    if (v17 == -1) break;
+    if (v17 > 128) goto LABEL_4030c1;
+    if (v17 > 82) { switch (v17) { case 83: ...; break; ... } }
+    else if (v17 == -131) { ... }
+    ...
+}
+```
+
+kuna **fails to recover this loop as a structured `while`**. Instead it:
+1. emits a bare label `label_2d00:` at the loop header (the `getopt_long` call),
+2. **hoists the entire `switch` out of the loop** to the very bottom of the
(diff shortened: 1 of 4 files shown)


mahaloz:

PR #50, opened 2026-06-25
[PROPOSAL] angr-optimized-memcpy-6301a9: recover unrolled-guard jump tables (MSVC optimized memcpy)

mahaloz:
# [PROPOSAL] optimized-memcpy-6301a9 — recover unrolled-guard jump tables (MSVC optimized memcpy)

**Status:** proposal (large feature — needs human go/no-go before implementation)
**angr testcase:** `test_decompiling_optimized_memcpy` :: `0x42cca0`
**Binary:** `i386/windows/736cb27201273f6c4f83da362c9595b50d12333362e02bc7a77dd327cc6b045a`
**Proposed option:** `unrolledguard on|off`

## The problem

On the MSVC CRT optimized memcpy (`0x42cca0`, Duff's device), angr recovers **7**
jump-table switches; kuna recovers **1** and degrades the other **6** to
`/* WARNING: Treating indirect jump as call */`. Full analysis + side-by-side:
[`analysis.md`](analysis.md) / [`angr-vs-kuna.txt`](angr-vs-kuna.txt).

The six failing tables share one root cause (see analysis): their bound comes from a
Duff's-device **unrolled/duplicated `cbranch` guard** at a block-merge point, and
kuna's `JumpBasic::check_unrolled_guard` (`s2_lift/jumptable.rs:2075`) is a documented
`// SEAM(structuring)` **stub**. The value-set / index-transform / table-emulation
path is already ported and works — only the unrolled-guard bound is missing.

## angr reference

angr's `JumpTableResolver` (in `angr/analyses/cfg`) resolves these as bounded jump
tables; the structurer then emits `switch` per table. The kuna-equivalent is Ghidra's
`JumpBasic::checkUnrolledGuard` / `checkCommonCbranch` (`jumptable.cc:1257`-ish) plus
the `BlockBasic` SSA helpers it depends on.

## Why this is LARGE (Hard Rule 7)

A decider subagent investigated and confirmed **large**:

- It is **not** a self-contained `kuna_*.rs` Action/Rule. The fix edits S2
  jumptable-recovery core (`analyzeGuards`/`findNormalized` in `jumptable.rs`) **beyond
  a single gated early-return**.
- It requires **new infrastructure**: porting the unported `BlockBasic` SSA helpers
  `findMultiequal` / `noInterveningStatement` / `unblockedMulti` / `liftVerifyUnroll`
  (all `// SEAM(W7)` in `substrate/block.rs`).
- It touches **>3 ported-core anchor files** (`jumptable.rs`, `block.rs`, `funcdata`,
  plus option/arch wiring) and is not cleanly gateable default-off (it changes core
  recovery, not an opt-in rendering pass).
- The rejected alternative — a `kuna_loweredswitch.rs`-style *manufacturing* Action
  that reads the absolute tables from the loadimage — is **even larger**: it duplicates
  working recovery, and the Duff's-device targets are mid-block, forcing CFG block
  splitting plus its own bounds inference.

1. **Port the W7 `BlockBasic` SSA helpers** `findMultiequal` / `noInterveningStatement`
   / `unblockedMulti` / `liftVerifyUnroll` in `substrate/block.rs` (faithful C++
   transcription; standard W7 SSA-helper effort).
2. **Port `checkCommonCbranch` + fill `check_unrolled_guard`** in `jumptable.rs` to
   detect the duplicated guard across the unrolled merge and contribute the bounding
   range constraint into `analyzeGuards`.
3. **Gate** behind a new arch flag + `option unrolledguard on|off`
   (`change_kind = structure-recovery`), default-off while developing; correct the
   stale `jumptable.rs` module-header SEAM note that misdescribes the chain as unported.
4. **Stage test** `tests/stages/ghangr-optimized-memcpy-6301a9.xml`: the bytechunk of
   `0x42cca0` + its in-`.text` tables; pass 1 (`option unrolledguard off`) asserts the
   `Treating indirect jump as call` degradation and `switch(` count == 1; pass 2
   (default/on) asserts the recovered switches (target: 7, matching angr).
5. **Ablation + speed**: `kuna test --all --baseline docs/baseline.json` must stay
   PARITY OK; measure decompile wall-time off vs on (jump-table recovery + extra SSA
   walks can be non-trivial; if over the speed budget, ship opt-in).

## Speed / risk assessment

- **Risk (correctness):** editing core jump-table recovery can regress the existing 1
  recovered switch and the broader datatest jump-table corpus. `make test` /
  `make test-stages` parity is the hard gate. Validate the bound across all six
  transformed tables (negative stride `-4`, offset `(a2-4)*4`, masks `&3`).
- **Risk (scope):** the W7 SSA helpers are reused well beyond this case, so the port
  should be done carefully/faithfully, not pattern-matched to this one function.
- **Speed:** the extra guard/SSA analysis runs only on functions with `size_in()>1`
  guard blocks; expected to be modest, but must be measured (Hard Rule 6).

## Recommended option name

`unrolledguard` (the construct is "jump table bounded by an unrolled/duplicated cbranch
guard"), `change_kind = structure-recovery`, `source_decompiler = angr`,
`inspiration = "test_decompiling_optimized_memcpy; JumpBasic::checkUnrolledGuard / angr JumpTableResolver; 0x42cca0"`.

---

### DIFF
--- a/docs/features/optimized-memcpy-6301a9/analysis.md
+++ b/docs/features/optimized-memcpy-6301a9/analysis.md
@@ -1,0 +1,90 @@
+# optimized-memcpy-6301a9 — analysis
+
+**angr testcase:** `test_decompiling_optimized_memcpy` :: `0x42cca0`
+**Binary:** `i386/windows/736cb27201273f6c4f83da362c9595b50d12333362e02bc7a77dd327cc6b045a`
+(an MSVC CRT `memcpy`/`memmove`, Duff's-device style)
+**angr:** 9.2.213
+
+## What angr does better
+
+The angr test asserts `d.codegen.text.count("switch") == 7`: angr recovers **all 7
+jump-table switches** in this optimized memcpy. The function is built from seven
+`jmp dword [reg*scale + const_table]` dispatches, where each table
+(`0x42cda8`, `0x42ce24`, `0x42cd28`, `0x42cfb0`, `0x42ceb4`, `0x42cf60`, …) holds
+*absolute* target addresses (PE has a fixed image base + relocations).
+
+kuna (driven on a self-contained bytechunk of the function at `0x42cca0`, since the
+i386 PE loader can't open the file directly — see *Loader note*) recovers **only 1**
+switch — the single plain-stride table `switch(a2 & 3)`. The other **6** indirect
+jumps fail recovery and degrade to:
+
+```c
+/* WARNING: Treating indirect jump as call */
+v1 = (void *)(**(code **)(v2 * 4 + 0x42cda8))();
+return v1;
+```
+
+i.e. `FlowInfo::truncateIndirectJump` (`s2_lift/flow.rs:2719`) converts the
+unrecovered `BRANCHIND` to a `CALLIND`. Side-by-side capture:
+[`angr-vs-kuna.txt`](angr-vs-kuna.txt).
+
+## The exact construct
+
+The one table that recovers is a plain `reg*4 + table`. The six that fail all have a
+**transformed switch index** *and* a guard that is duplicated across the unrolled
+Duff's-device ladder's merge points:
+
+| dispatch | index transform |
+|---|---|
+| `v2 * -4 + 0x42cf60`        | negative stride |
+| `(a2 - 4) * 4 + 0x42ce24`   | offset |
+| `((uint4)v4 & 3) * 4 + 0x42ceb4` | mask |
+| `(a2 & 3) * 4 + 0x42cfb0`   | mask |
+| `(v2 * 4) + 0x42cda8`       | plain, but guarded at an unrolled merge |
+| `((uint4)a0 & 3) * 4 + 0x42cd28` | mask |
+
+## Owning stage and root cause
+
+**Stage:** S2 lift, switch-model sub-stage — jump-table recovery
+(`decompiler/crates/kuna-decomp/src/s2_lift/jumptable.rs`).
+
+The index transforms themselves are **not** the blocker: `JumpBasicModel`'s
+`find_unnormalized_basic` / `circlerange_pull_back` (a real `CircleRange` value-set)
+/ `build_addresses_basic` (a live `EmulateFunction`) are all ported and already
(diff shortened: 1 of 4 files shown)


mahaloz:
## Closing — implemented by #105 (`unrolledguard`)

Re-triaged against today's `main` by running kuna on `0x42cca0` in the MSVC CRT binary and ablating.

**The loader note is stale** — the i386 PE opens directly today, no bytechunk harness needed.

**All 7 tables recover.**

| | **defaults today** | **`--option unrolledguard on`** | angr 9.2.213 |
|---|---|---|---|
| `switch` tokens | 1 | **16** | 7 |
| `Treating indirect jump as call` | 7 | **0** | 0 |
| loc | 164 | 462 | — |
| wall time | 2.3 s | 4.2 s | — |

The defaults column reproduces this proposal's baseline exactly (1 recovered switch, the rest degraded). With the option on, every one of the seven dispatch tables named in the degraded output is recovered: `0x42ceb4`, `0x42cfb0` (×2), `0x42cf60`, `0x42cda8`, `0x42cd28`, `0x42ce24`.

The `switch`-token count is 16 rather than angr's literal 7 because kuna emits one `switch` per *structural position* of the same seven Duff's-device dispatches — this is not tail duplication (the count is unchanged at 16/0 with `taildup` and `dedupitetail` both forced off), and the emitted shape matches angr's reference: `switch(a2) { case 0..3: goto label_42cfc0/c8/d8/ec; default: <nested switch on & 3> }`.

**Default-off is what this proposal sanctioned.** `unrolledguard` ships default-off / `destructive_as_default`. Plan step 5 of this proposal states verbatim: *"measure decompile wall-time off vs on … if over the speed budget, ship opt-in."* Measured 2.3 s → 4.2 s on this function — over the ≤5% budget, so opt-in is the disposition the proposal itself specified. #105 also delivered the two-pass stage test the plan asked for (pass 1 asserts the degradation, pass 2 asserts the recovery).

**Residual (cosmetic):** the `switch`-token count differs from angr's literal 7 for the structural-copy reason above; the same seven tables are recovered either way.

PR #53, opened 2026-06-25
[PROPOSAL] angr-true-1804-x86-64-8a690c: condition-aware if/else tail dedup (S8 structuring)

mahaloz:
# [PROPOSAL] `dedup-ite-tail` — condition-aware if/else tail deduplication (S8 structuring)

**Opportunity:** `test_decompiling_true_1804_x86_64::usage`
**Binary:** `/home/mahaloz/github/angr-dev/binaries/tests/x86_64/true_ubuntu1804` — `usage` (x86_64 @ `0x401610`)
**angr:** 9.2.213 · **Scope:** LARGE (Hard rule 7) · **Proposed option:** `dedup-ite-tail` (default-off) · **Proposed ElementId:** 4100

## The problem

On `usage`, angr's decompilation is materially cleaner than kuna's
(loc 112 vs 121, gotos 2 vs 3, labels 2 vs 3, ifs 6 vs 9). The full analysis is
in [`analysis.md`](./analysis.md); the side-by-side is in
[`angr-vs-kuna.txt`](./angr-vs-kuna.txt). Two structuring-quality differences:

1. **Duplicated if/else tail (dominant).** kuna's trailing
   `if (v28 == 0) { A } else { B }` duplicates a shared prefix and suffix across
   *both* arms — the `__printf_chk("...online help...")` call, the `setlocale(5,0)`
   call, the `strncmp(...,"en_",3)` check, and the `__printf_chk("...Full
   documentation...")` call all appear twice. angr emits each once and converges the
   two distinct continuations through shared labels (`LABEL_4018e8`/`LABEL_401822`).
2. **Loop shape.** kuna emits the inlined `rep cmps` loop with a `goto` out of the
   loop body; angr renders it as a `do/while` terminated by a `break`.

Both are decisions of the **structurer**, i.e. kuna's S7/S8 `blockaction`
CollapseStructure/TraceDAG region-collapse engine and its goto-vs-`break` schema
selection (`decompiler/crates/kuna-decomp/src/s8_structure/blockaction.rs`).

## Why this is not a single gated Action (the scope call)

A decider subagent (verbatim verdict recorded in `record.json → decisions`)
returned **`scope: large`, `proposal_required: true`**. Reasoning:

- The duplication is produced by the structurer **cloning a shared continuation**
  instead of converging two predecessors onto one labelled block. There is no
  pre-structuring dataflow form that "shares" the tail — sharing it is a
  condition-aware ITE region-dedup capability that lives *inside*
  CollapseStructure/TraceDAG.
- Unlike `kuna_loweredswitch` (which fabricates an S2 `BRANCHIND`+`JumpTable` and
  lets the **unchanged** structurer render it), there is **no manufacturable upstream
  CFG artifact** that makes Ghidra's collapse engine share the continuation.
- Closing the gap therefore requires a new pass *type* touching S8
  `blockaction` region collapse — exactly Hard rule 7's "new infrastructure" and
  "touch S7/S8 structuring beyond a single gated early-return" triggers.

## angr reference

- angr's structurer condition-processing / ITE region handling — the
  `RegionSimplifier` / condition-deduplication path in
  `angr/analyses/decompiler/region_simplifiers/` and the structurer's
  `_make_ites` / condition-node merging that converges shared continuations onto
  labelled successors rather than cloning them. (kuna already has the analysis-only
  region tree from the RegionIdentifier port: `kuna_regionid` / `kuna_regiongraph`,
  S7 — a natural substrate.)

## Proposed implementation plan (multi-step — for human go/no-go)

1. **Detection (S7, analysis-only).** Using the existing `kuna_regionid` region
   tree, identify an `if/else` region whose two arms share a maximal common
   *prefix* and/or *suffix* of statement-equivalent blocks (same PcodeOp sequence,
   modulo the condition varnode). Bound it (small fixed cap on hoisted blocks) to
   keep it safe and fast.
2. **Transform (S8, gated).** Behind `option dedup-ite-tail` (default-off), rewrite
   the region so the shared prefix/suffix is emitted once and the divergent middles
   are reached via labelled blocks — i.e. teach the collapse engine (or a
   pre-collapse normalizer feeding it) to converge the predecessors. This is the
   genuinely new region-collapse logic.
3. **Loop schema (optional, separable).** A follow-up gated tweak to
   goto-vs-`break` selection for the `rep cmps` loop; can ship separately.
4. **Verify & gate.** New `tests/stages/ghangr-true-1804-x86-64-8a690c.xml`
   (off=current duplicated tail, on=deduplicated). `kuna test --all` ablation must
   stay PARITY OK with the option default-off; measure target decompile speed
   off-vs-on (structurer changes are the risk).

## Speed / risk assessment

- **Risk: HIGH.** Region-collapse changes are the most correctness-sensitive part of
  the pipeline — a wrong convergence emits invalid C. Must stay default-off until an
  ablation over the full datatest corpus is clean.
- **Speed:** detection is O(region · block-eq) over a bounded window; expected
  small, but must be measured per Hard rule 6 because it runs in the structurer.
- **Generality:** the shared-prefix/suffix-across-if-arms shape is common in
  gettext/`--help`-style code (coreutils family) and in canary/cleanup epilogues, so
  the pass should generalize beyond this single function — but that breadth is also
  why it must be gated and corpus-ablated before any default-on consideration.

## Recommendation

Approve as a **draft proposal**. On go, re-dispatch an implementation worker on this
branch to build the S7 detector + S8 gated region converter as `dedup-ite-tail`
(ElementId 4100, default-off), with the stage test and full ablation.

### DIFF
--- a/docs/features/true-1804-x86-64-8a690c/analysis.md
+++ b/docs/features/true-1804-x86-64-8a690c/analysis.md
@@ -1,0 +1,101 @@
+# Analysis — `test_decompiling_true_1804_x86_64 :: usage`
+
+- **Opportunity:** `test_decompiling_true_1804_x86_64::usage`
+- **Binary:** `/home/mahaloz/github/angr-dev/binaries/tests/x86_64/true_ubuntu1804`
+- **Function:** `usage` (x86_64 @ `0x401610`)
+- **angr version:** 9.2.213
+- **Owning stage:** S7/S8 — `blockaction` (Stage 16, *Structuring*): the
+  CollapseStructure/TraceDAG region-collapse engine + ConditionalExecution
+  (`condexe`) + goto/schema selection.
+
+## What angr does better
+
+The full side-by-side is in `angr-vs-kuna.txt`. Metrics (reference | kuna):
+
+| metric | angr | kuna |
+|---|---|---|
+| loc | 112 | 121 |
+| gotos | 2 | 3 |
+| labels | 2 | 3 |
+| ifs | 6 | 9 |
+| loops | 2 | 2 |
+
+The pipeline "recovery-failure marker" signal here is a **false positive** — it
+fires on the `/* WARNING: Subroutine does not return */` comment kuna emits before
+the tail-call `exit(a0)`, which is a normal Ghidra no-return annotation, not a
+structuring abort. The real gap is two structuring-quality differences:
+
+### 1. Duplicated if/else tail (the dominant difference)
+
+kuna's trailing block is:
+
+```c
+label_17b8:
+  v4 = *(int8 *)&v6[0x18];          // angr: v28
+  if (v4 == 0) {
+    __printf_chk(1, "...online help...", ...);   // (a)
+    v3 = setlocale(5,0);                          // (b)
+    if (v3 != 0) {
+      v1 = strncmp(v3,"en_",3);                   // (c)
+      if (v1 != 0) { v4 = 0x4764; goto label_18e8; }
+    }
+    v4 = 0x4764; v3 = " invocation";
+    __printf_chk(1, "...Full documentation...", ...);   // (d)
+  }
+  else {
+    __printf_chk(1, "...online help...", ...);   // (a)  DUPLICATE
+    v3 = setlocale(5,0);                          // (b)  DUPLICATE
+    if (v3 != 0) {
+      v1 = strncmp(v3,"en_",3);                   // (c)  DUPLICATE
+      if (v1 != 0) {
+label_18e8:
+        __printf_chk(1, "...Report translation bugs...", ...);
+      }
+    }
+    __printf_chk(1, "...Full documentation...", ...);   // (d)  DUPLICATE
+    v3 = ""; if (v4 == 0x4764) v3 = " invocation";
+  }
+```
+
+The online-help printf `(a)`, the `setlocale` `(b)`, the `strncmp` `(c)` and the
+Full-documentation printf `(d)` are emitted **twice** — once in each arm of
+`if (v28 == 0)`. angr emits each of them **once** and threads the two genuinely
+distinct continuations through shared labels:
+
+```c
+    if (v28) { ...; goto LABEL_401822_or_4018e8; }
+    else     { ...; LABEL_4018e8: report_bugs; LABEL_401822: full_documentation; }
+```
+
+This is angr's structurer doing condition-aware **region deduplication** (its
+"ITE region converter" / condition-processing): two predecessors are converged onto
+one labelled successor instead of cloning the shared continuation. kuna inherits
+Ghidra's collapse-based structurer, which clones the shared tail into both arms — so
+it pays an extra `if`, extra lines, and one extra goto/label.
+
+### 2. Loop shape
+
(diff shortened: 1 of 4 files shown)


mahaloz:
## Closing — target metrics now beat angr; the proposal's dominant premise doesn't hold up against its own reference

Re-triaged against today's `main` by running kuna on `true_ubuntu1804::usage` and ablating all 14 P8 goto-quality options plus both `--mode` presets.

**Method validated first.** Applying the counters to the branch's own stored text reproduces this PR's metric table exactly (angr 112/2/2/6, old-kuna 121/3/3/9), so the numbers below are comparable.

|---|---|---|---|
| loc (non-blank) | 121 | **108** | 112 |
| gotos | 3 | **1** | 2 |
| labels | 3 | **1** | 2 |
| ifs | 9 | 9 | 6 |
| loops | 2 | 2 | 2 |

kuna now **beats** angr on loc, gotos and labels, ties on loops, and trails only on ifs.

**Sub-item (2) — loop shape — is fixed.** The inlined `rep cmps` loop is now `do { if (v3 == 0) break; … } while (v12);` with three `break;` statements and zero gotos, instead of `goto label_17af` / `goto label_17b8`. That alone accounts for the 3→1 goto and 3→1 label drop.

**Sub-item (1) — the "dominant" duplication — was never real.** This PR states the online-help `__printf_chk`, `setlocale(5,0)`, `strncmp(…,"en_",3)` and Full-documentation `__printf_chk` *"all appear twice [in kuna]. angr emits each once."* Grepping the angr pane of this branch's own `angr-vs-kuna.txt`:

| | today kuna | angr ref | old kuna |
|---|---|---|---|
| `online help` | 2 | **2** | 2 |
| `setlocale(` | 2 | **2** | 2 |
| `strncmp(` | 2 | **2** | 2 |
| `Full documentation` | 2 | **2** | 2 |

angr duplicates all four too. `analysis.md` elides this — its angr paraphrase writes each arm's duplicated prefix as `…`, hiding that angr clones the same three statements. angr's only real edge was threading two labels + two gotos; kuna today does the mirror-image convergence with one label + one goto.

**Note on `dedupitetail` (#104).** #104 was titled for this function, but on this target it is a byte-identical no-op today (`--option dedupitetail off` diffs empty), and #104's own body says so: *"the proposal's target true_ubuntu1804::usage has its duplication sub-block-entangled … fully matching angr there needs the deeper label-convergence transform."* The improvement here came from unflagged bug-fix-tier structuring/emit work instead — leading candidate #150 (`markLabelBumpUp`). `dedupitetail` still earns its keep on the general leaf-hoist case; it just isn't what moved this function.

**Residual → tracked elsewhere.** The only remaining metric is ifs, 9 vs 6, and none of it is if/else tail duplication:

- **+2** — angr short-circuits each guard pair into one compound condition (`if (!v30 || !strncmp(v30, "en_", 3))`), where kuna emits two nested ifs per arm, twice.
- **+1** — kuna's trailing `if (v3 == 0x4764) { v2 = " invocation"; }`, which angr constant-folds to a plain assignment.

The +2 is short-circuit condition folding — the same root as #39, and it belongs there rather than in a tail-dedup PR.

PR #46, opened 2026-06-25
[PROPOSAL] angr-fmt0-main-85c389: structure irreducible (multi-entry) loops

mahaloz:
# [PROPOSAL] angr-fmt0-main-85c389: structure irreducible (multi-entry) loops

**Status:** draft proposal — awaiting human go/no-go. **Do not implement before approval.**

- **Opportunity:** `test_decompiling_fmt0_main :: main`
- **Binary:** `/home/mahaloz/github/angr-dev/binaries/tests/x86_64/fmt_0` (GNU coreutils `fmt`), `main` @ `0x4019b0`, x86_64
- **Reference:** angr 9.2.213
- **Proposed option:** `irreducibleloops` (change_kind: `structure-recovery`)
- **Scope:** **large** (scope-decider verdict in `record.json`)

## The problem

kuna emits **syntactically invalid C** for `fmt`'s `main` because `main` contains an
**irreducible (multi-entry) loop**: the `-WIDTH` first-arg special case `goto`s into the
`getopt_long` option-parsing loop body. kuna's Ghidra collapse-based structurer resolves
irreducibility only by emitting gotos, and additionally produces a while-block whose
condition carries side-effecting multi-statements that the S9 printer renders as incomplete
`while (...` lines. angr recovers a clean, valid `while(true){ ...; if(v==-1) break; ... }`.

Full evidence: [`analysis.md`](analysis.md), side-by-side: [`angr-vs-kuna.txt`](angr-vs-kuna.txt).

## angr reference

- `angr/analyses/decompiler/region_identifier.py` — `RegionIdentifier` (already ported,
  read-only, as kuna `kuna_regionid`/`kuna_regiongraph` under `s7_regions/`).
- `angr/analyses/decompiler/structuring/` — the SAILR/Phoenix condition-based structurer
  (`PhoenixStructurer`) with `_refine_cyclic` / goto-and-loop refinement that handles
  multi-entry loops by node duplication / condition recovery. **Not yet ported.**

## Why this is not a single Action/Rule

`loweredswitch` works by **manufacturing a missing S2 `JumpTable`** that the existing
structurer already consumes. Here the structurer *itself* is wrong on irreducible loops —
there is no artifact to pre-build. Closing the gap requires new structuring infrastructure
across S7 (region structuring), S8 (`blockaction.rs`), and S9 (`printc.rs`), > 3 anchor
files and > 1 new module. The decider verdict (verbatim in `record.json`):

[...]

PR #54, opened 2026-06-25
[PROPOSAL] angr-morton-my-message-callback-bfd2fa: SAILR-style goto-reduction (tail duplication)

mahaloz:
# [PROPOSAL] angr-morton-my-message-callback: SAILR-style goto-reduction by epilogue tail-duplication

**Scope:** large (per Hard rule 7 and the recorded decider verdict).

## The problem

On `my_message_callback` in `morton` (x86_64, @ `0x403de7`) the **only** structural
difference between angr and kuna is a single residual `goto` + `label:`:

| metric | angr | kuna |
|---|---|---|
| gotos | **0** | **1** |
| labels | **0** | **1** |
| ifs | 9 | 9 |
| loops / switches | 0 | 0 |

kuna:

```c
        else {
          v1 = strcmp(v3,"UV");
          if (v1 != 0) {
            v5 = strlen("Invalid Command");
            mosquitto_publish(a0,0,"app/error",v5 & 0xffffffff,"Invalid Command",1);
            goto label_41d6;
          }
          ...
        }
      ...
label_41d6:
  free(v6);
  return;
```

angr (zero gotos) — it **duplicates** the shared `free(ptr); return;` epilogue into each
arm that reaches it, so the cross-edge becomes a structured early return:

```c
        else
        {
            v5 = mosquitto_publish(a0, 0, "app/error", strlen(v9) & 4294967295, v9, 1, 0);
            free(ptr);
            return;
        }
        puts(ptr);
        ...
```

See `analysis.md` and `angr-vs-kuna.txt` for the full side-by-side.

## The angr reference pass / class

This is angr's **goto-reduction via tail duplication** (the Phoenix/SAILR family in
`angr.analyses.decompiler` — the region-simplification + `DuplicationOptimizer` /
`GotoSimplifier` style transforms). angr's structurer is *allowed* to duplicate small,
side-effect-bounded tails (a `return`, a `free(x); return`) to eliminate an unstructured
edge; its quality metric explicitly trades a few duplicated statements for zero gotos.

## Why this is large (owning stage: S8 structuring)

The goto is produced by `decompiler/crates/kuna-decomp/src/s8_structure/blockaction.rs`,
the **verbatim port of Ghidra's `CollapseStructure` / `TraceDAG` / `BlockTrace`**. Its
header states the schema-precedence order and the DAG bad-edge scoring are
"output-determining ... transcribed verbatim" — they *choose* which edge becomes a goto.
Ghidra has **no schema that duplicates a tail to avoid a goto**; that transform simply
does not exist in the collapse algorithm.

Matching angr therefore requires a **new structuring transform**, not a gated `Action`
that manufactures an artifact for the existing structurer (the way `kuna_loweredswitch.rs`
manufactures an S2 BRANCHIND/JumpTable). Concretely it must:

1. identify unstructured ("goto") edges after `CollapseStructure` whose **target** is a
   small epilogue block (statement-count ≤ threshold, no incoming side effects that would
   change semantics if duplicated, e.g. a `free(p); return;`);
2. **duplicate** that target's statements into the goto source (a CFG/`BlockGraph` mutation
   — clone the `BlockBasic`, its ops, and rewire edges), then drop the unstructured edge;
3. re-run / locally re-collapse the structuring so the duplicated tail folds into the arm
   as a structured early return;
4. preserve def/use and ordering (the duplicated `free`/`return` are real p-code ops, so
   this is *not* a pure print-time rewrite).

That is a **new pass type touching S8 structuring/region code beyond a single gated
early-return** — two independent triggers of Hard rule 7. The decider verdict (recorded in
`record.json`) is `scope: large`.

## Proposed implementation plan (multi-step — for the approved worker)

- **New module** `kuna_gotoreduce.rs` (S8) — a post-`CollapseStructure` `Action`,
  option-gated, default-OFF.
- **Anchor edits (minimal, `// (kuna)`):** register the action in `universalaction.rs` /
  the relevant `coreaction*` registration; option flag on the architecture struct
  (reset-default off); `options.rs`; `stages.toml` `settableTable` row.
- **Core mechanics:** add a `BlockGraph` helper to clone a `BlockCopy`/`BlockBasic` tail
  and rewire the floating edge (this is the genuinely new, risky piece — it mutates the
  structured block tree the verbatim engine produced).
- **Bounds / fail-safe:** cap duplicated statement count (e.g. ≤ 4 ops, single-exit,
  ends in `return`), bail on any block with multiple non-goto predecessors that would
  change semantics, mirroring loweredswitch's fail-safe cap.
- **Tests:** `tests/stages/ghangr-morton-my-message-callback-bfd2fa.xml` two-pass
  (off=goto present, on=goto gone); plus a synthetic minimal CFG fixture.

## Speed / risk assessment

- **Risk: high.** This mutates the structured block tree after a verbatim-ported engine.
  Incorrect duplication can change semantics (duplicating a block with side effects reached
  by more than the goto source) or desync def/use. The bounds above contain it but need
  careful adversarial testing across the datatest corpus.
- **Ablation risk:** likely non-zero — tail-duplication changes rendering on many
  functions that currently use a clean shared-epilogue goto, so this is expected to ship
  **default-OFF opt-in** even if correct.
- **Speed:** an extra post-structuring scan + occasional re-collapse; expected small but
  must be measured against the +5% budget on the target.
- **Same family** as the irreducible-loop SAILR gap (proposal PR #46) — worth coordinating
  so kuna grows one coherent goto-reduction layer rather than point fixes.

## Proposed option name

`gotoreduce` (`change_kind = structure-recovery`, `source_decompiler = angr`,
`inspiration = test_decompiling_morton_my_message_callback; SAILR/Phoenix goto-reduction (tail duplication); my_message_callback`).

## Recommendation

**Go** if the team wants kuna to start matching angr/SAILR on goto count (high-value,
broadly applicable, but real structuring risk) — staff it as a deliberate S8 effort with a
full adversarial test pass. **No-go / defer** if structuring-quality parity with angr is
out of scope right now; the gap is cosmetic (one goto, semantics already correct) and the
implementation cost/risk is disproportionate to this single function.

### DIFF
--- a/decompiler/docs/features/morton-my-message-callback-bfd2fa/angr-vs-kuna.txt
+++ b/decompiler/docs/features/morton-my-message-callback-bfd2fa/angr-vs-kuna.txt
@@ -1,0 +1,191 @@
+==============================================================================
+test_decompiling_morton_my_message_callback :: my_message_callback   (angr 9.2.213 @ 0x403de7)
+==============================================================================
+
+--- angr ---
+typedef struct struct_0 {
+    char padding_0[8];
+    long long field_8;
+    char *field_10;
+    char padding_18[8];
+    char field_20;
+} struct_0;
+
+extern char *cfg;
+extern char g_40d00a;
+extern char g_41316f;
+extern long long last_mid;
+extern char process_messages;
+
+void my_message_callback(long long a0, unsigned long long a1, struct_0 *a2, unsigned long long a3)
+{
+    unsigned long len;  // rax
+    unsigned long len;  // rax
+    unsigned long len;  // rax
+    unsigned long long v0;  // [bp-0x78]
+    unsigned long long v1;  // [bp-0x68]
+    unsigned int v2;  // [bp-0x50]
+    unsigned int v3;  // [bp-0x4c]
+    unsigned int v4;  // [bp-0x48]
+    unsigned int v5;  // [bp-0x44]
+    char *tok;  // [bp-0x40]
+    unsigned long tok;  // [bp-0x38]
+    unsigned long v8;  // [bp-0x30]
+    unsigned long v9;  // [bp-0x28]
+    unsigned long ptr;  // [bp-0x20]
+
+    v1 = a1;
+    v0 = a3;
+    tok = strtok(a2->field_10, " ");
+    tok = strtok(NULL, &g_40d00a);
+    v8 = "Invalid Topic";
+    v9 = "Invalid Command";
+    ptr = malloc(strlen(cfg) + 44);
+    v2 = 0;
+    v3 = 0;
+    v4 = 0;
+    v5 = 0;
+    if (process_messages != 1)
+    {
+        free(ptr);
+        return;
+    }
+    if (g_41316f && a2->field_20)
+        mosquitto_publish(a0, &last_mid, a2->field_8, 0, 0, 1, 1);
+    if (tok && !mosquitto_pub_topic_check(tok))
+    {
+        if (!tok)
+        {
+            v5 = mosquitto_publish(a0, 0, "app/error", strlen(v9) & 4294967295, v9, 1, 0);
+            free(ptr);
+            return;
+        }
+        if (!strcmp(tok, "temperature"))
+        {
+            v2 = generate_random(68, 82);
+            len = strlen(cfg);
+            snprintf(ptr, len + 43, "%s recorded a temperature of %d degrees", cfg, v2);
+        }
+        else if (!strcmp(tok, "humidity"))
+        {
+            v3 = generate_random(50, 100);
+            len = strlen(cfg);
+            snprintf(ptr, len + 34, "%s recorded a humidity of %d%%", cfg, v3);
+        }
+        else if (!strcmp(tok, "UV"))
+        {
+            v4 = generate_random(0, 11);
+            len = strlen(cfg);
+            snprintf(ptr, len + 32, "%s recorded a UV index of %d", cfg, v4);
+        }
+        else
+        {
+            v5 = mosquitto_publish(a0, 0, "app/error", strlen(v9) & 4294967295, v9, 1, 0);
+            free(ptr);
+            return;
+        }
+        puts(ptr);
+        v5 = mosquitto_publish(a0, 0, tok, strlen(ptr) & 4294967295, ptr, 1, 0);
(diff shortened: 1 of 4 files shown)


mahaloz:
## Closing — implemented by #103 (`taildup`), default-on since #110

Re-triaged against today's `main` by running kuna on `my_message_callback` and ablating.

**Criterion met exactly.** This proposal's success condition was *"the single concrete structural difference is one residual `goto` + its `label:`"* → drive it to angr's 0/0.

|---|---|---|---|
| gotos | 1 | **0** | 0 |
| labels | 1 | **0** | 0 |
| ifs | 9 | 9 | 9 |
| switches | 0 | 0 | 0 |

The exact construct is gone. The `strcmp(v2,"UV") != 0` arm now ends with an inlined `free(v5); return;` early return carrying an explicit marker, and there is no `label_41d6:` anywhere in the output:

```c
/* WARNING: taildup: duplicated return-call tail to remove goto */
```

**Ablation isolates the responsible feature.** `--option taildup off` reproduces this proposal's baseline verbatim — **1 goto (`goto label_41d6;`) / 1 label (`label_41d6:`) / 9 ifs**. `--option gotoreduce off` alone changes nothing (still 0/0); `taildup off + gotoreduce off` gives 1/1. So `taildup` — not the `gotoreduce` name this proposal suggested — is the option that actually closes it.

**Why #103 is this proposal's implementation.** #103 ("taildup: SAILR tail duplication goto reduction") implements exactly the transform specified here: duplicate a small, bounded, side-effect-limited return tail into the goto source so the cross-edge becomes a structured early return. It mirrors angr `ReturnDuplicatorBase`, whose budget is calls (`max_calls_in_regions=2`) rather than call-rejection — which is precisely why it handles the `free(p); return;` epilogue that `gotoreduce` (rejects all calls) and `crossjumprevert` (non-return fall-through tails) both decline. `docs/options.md` records the win verbatim: *"morton my_message_callback: 1 goto -> 0"*. The bounds this proposal asked for shipped too (≤3 blocks / ≤12 printed ops / ≤2 calls, no STORE).

One deviation worth recording: it landed **default-on** (DIV-14, #110), not the default-off opt-in this proposal expected. That was decided on ablation evidence, and `docs/options.md` documents the flip-off case (many early-exit error checks sharing one cleanup epilogue — the C `goto out;` idiom).

PR #56, opened 2026-06-25
[PROPOSAL] angr-newburry-main-57ccb1: SUID-guard goto is the SAILR goto-reduction family (large)

mahaloz:
# [PROPOSAL] angr-newburry-main-57ccb1 — eliminate the SUID-guard goto via condition-based join restructuring

**Status:** draft proposal — needs human go/no-go before any implementation worker is spent.

**Opportunity:** `test_decompiling_newburry_main :: main`
(`/home/mahaloz/github/angr-dev/binaries/tests/x86_64/decompiler/newbury`, lighttpd `server.c::main`, x86_64 @ `0x40f696`).

**Proposed option name:** `gotoreduce` (shared with the sibling proposal — see below; do **not** mint a per-binary option).

## The problem

kuna's `newburry/main` differs from angr's by exactly **one goto + one label**
(all other metrics match: 6 loops, ~20 ifs, 1 ternary). The goto is a forward
jump *into* a then-branch:

```c
  if (getuid() == 0) {
label_f6ec:  setlocale(...); ... do { server loop } while(...);   // BODY
  } else {
    ... if (egid==gid) goto label_f6ec;                            // <-- goto into BODY
    v2 = -1; fputs("Are you nuts ...", stderr);
  }
  return v2;
```

angr emits zero gotos by **inverting the SUID-check guards** so the exceptional
`fputs; return -1` path is nested and the shared **BODY** (`setlocale; ...; server
loop`) becomes the fall-through join:

```c
  if (getuid()) {
      if (geteuid() != getuid() || !(getegid() == getgid())) {
          fputs("Are you nuts ...", stderr); return -1;
      }
  }
  setlocale(...); while (true) { server loop }                     // BODY = fall-through
```

Full side-by-side: [`angr-vs-kuna.txt`](./angr-vs-kuna.txt). Root-cause analysis:
[`analysis.md`](./analysis.md).

## Why it is LARGE (Hard rule 7)

The goto is produced in **S8 structuring**,
`decompiler/crates/kuna-decomp/src/s8_structure/blockaction.rs` — a verbatim port
of Ghidra's `CollapseStructure`/`TraceDAG`. Reproducing angr's output requires
**condition-based block-graph restructuring**: invert the guard predicates so the
exceptional path nests, then re-parent the multi-predecessor BODY out of the first
`if`-then to become the fall-through join. That is a **new pass type** mutating the
BlockGraph, not a value-level Action/Rule and not a gated early-return:

- A gated early-return in `blockaction.rs` can only *suppress* the goto fallback;
  it cannot *synthesize* a guard inversion or re-parent a join — so it would yield
  invalid/garbled C, not angr's structure.
- The canonical small template (`kuna_loweredswitch.rs`) is an **S2 pcode** Action,
  inapplicable to block-graph structuring.

This is the **same root** as two already-filed proposals:

- **`kuna-gotoreduce-gap`** — morton/my_message_callback, **PR #54**: 1 goto+label
  to a shared *epilogue*; angr *tail-duplicates*. Ruled large (new S8 pass type).
- **`kuna-irreducible-loop-gap`** — fmt0/main family, **PR #46**: needs promoting
  kuna's read-only `RegionIdentifier` (`s7_regions/kuna_regionid.rs`) into a real
  **emit-path SAILR/Phoenix condition-based structurer**.

The pipeline record states this whole goto-reduction family shares **one** root and
should be closed by **one separately-funded project**, not point fixes. This
`newburry/main` case is another member: it is a *join-fall-through via guard
inversion* variant (vs. morton's *tail-duplication* variant), but both are emitted
by the same `CollapseStructure` fallback and both need the same SAILR emit-path
structurer.

## Proposed implementation plan (when the family project is funded)

This is **not** a standalone implementation — it folds into the shared SAILR
emit-path structurer that closes `gotoreduce` (PR #54) and `irreducibleloops`
(PR #46). The pieces specific to this case:

1. **Promote `s7_regions/kuna_regionid.rs`** from a read-only console surface into
   an emit-driving structurer (the shared family prerequisite).
2. Add a **condition-based goto-elimination** transform over the region/block
   graph that recognizes a multi-predecessor join `J` reached by (a) a then-branch
   and (b) a forward goto from a sibling branch, where the non-goto remainder of
   the sibling is a *terminating* path (returns/exits). Transform: invert the
   sibling's guard chain (de Morgan) so the terminating path nests, hoist `J` to be
   the fall-through join after the merged guard, and drop the goto+label.
3. Gate the whole structurer behind the **`gotoreduce`** option (default OFF while
   developing), early-returning to byte-identical output when off.
4. Verify on the family corpus (morton, newburry, and the irreducible-loop set)
   under one option to avoid per-binary option sprawl.

## Speed / risk assessment

- **Risk:** high — block-graph restructuring is the most semantically delicate part
  of the pipeline; an incorrect guard inversion produces *wrong* C, not just ugly C.
  Must be guarded by strict shape preconditions (terminating sibling remainder,
  single join, no intervening side effects on the inverted predicate evaluation
  order) and validated against the full 675-assertion datatest corpus.
- **Speed:** an S8 graph pass over already-collapsed structure; expected within the
  +5% budget, but must be measured on the family corpus (newburry/main is a large
  function — good speed canary). Not measured in this proposal (no implementation).
- **Default:** even if the family ablation is clean, this stays **opt-in** until the
  shared structurer is proven across the whole family.

## Recommendation

Do **not** dispatch a standalone implementation worker for `newburry/main`. Fold it
into the funded **`gotoreduce` / SAILR emit-path structurer** project (PR #54 / PR
#46). Track this binary as a *regression target* for that project.

### DIFF
--- a/docs/features/newburry-main-57ccb1/analysis.md
+++ b/docs/features/newburry-main-57ccb1/analysis.md
@@ -1,0 +1,112 @@
+# newburry/main — angr-vs-kuna analysis (newburry-main-57ccb1)
+
+- **angr testcase:** `test_decompiling_newburry_main :: main`
+- **Binary:** `/home/mahaloz/github/angr-dev/binaries/tests/x86_64/decompiler/newbury` (lighttpd `server.c::main`)
+- **angr version:** 9.2.213, function @ `0x40f696`
+- **Selector:** `main` (arch x86_64)
+
+## The gap (one concrete structural difference)
+
+Metrics (reference = angr | kuna):
+
+| metric | angr | kuna |
+|---|---|---|
+| gotos | **0** | **1** |
+| labels | **0** | **1** |
+| ifs | 20 | 16 |
+| loops | 6 | 6 |
+| ternaries | 1 | 1 |
+
+The **only** structural defect is a single `goto` + `label`. Everything else
+matches (both 6 loops, ~equal if/ternary counts). kuna's output is otherwise
+clean and correct.
+
+### kuna (abridged)
+
+```c
+  v2 = getuid();
+  if (v2 == 0) {
+label_f6ec:
+    setlocale(2,0x43620);
+    tzset();
+    do { ... big body / server loop ... } while (dat_5f278 != 0);   // BODY
+  }
+  else {
+    v2 = geteuid();
+    v3 = getuid();
+    if (v2 == v3) {
+      v2 = getegid();
+      v3 = getgid();
+      if (v2 == v3) goto label_f6ec;       // <-- forward goto INTO the then-branch's BODY
+    }
+    v2 = -1;
+    fputs("Are you nuts ? Don\'t apply a SUID bit to this binary\n",dat_5f140);
+  }
+  return v2;
+```
+
+### angr (abridged)
+
+```c
+  if (getuid()) {
+      v1 = geteuid();
+      if (v1 != getuid() || !(getegid() == getgid())) {
+          fputs("Are you nuts ? ...", stderr);
+          return 4294967295;               // exceptional path nested + early-return
+      }
+  }
+  setlocale(2, "C");                        // BODY is the fall-through JOIN
+  tzset();
+  while (true) { ... big body / server loop ... }
+```
+
+## What angr does better
+
+The SUID self-check has a body region **BODY** (`setlocale; tzset; <server loop>`)
+that is reached from **two predecessors**:
+
+1. the `getuid() == 0` path, and
+2. the "euid==uid && egid==gid all pass" path.
+
+- **kuna** (verbatim Ghidra `CollapseStructure`/`TraceDAG`) nests BODY inside the
+  first `if`-then and reaches it from the second path via a **forward `goto` into
+  that branch** (`goto label_f6ec`). Syntactically valid, but a goto + label.
+- **angr** (SAILR/Phoenix condition-based structurer) **inverts the guard
+  conditions** so the *exceptional* `fputs; return -1` path becomes the nested
+  branch, and **BODY becomes the fall-through join** after the guard. No goto, no
+  label.
+
+## Owning stage
+
+- **S8 structuring** — `decompiler/crates/kuna-decomp/src/s8_structure/blockaction.rs`
+  (verbatim port of Ghidra `CollapseStructure` / `TraceDAG`). The goto is emitted
+  here when the collapse cannot find a structured form for a multi-predecessor
+  join, so it falls back to a goto edge.
+- The would-be fix layer is **S7 region identification**
+  (`s7_regions/kuna_regionid.rs`), kuna's port of angr's `RegionIdentifier`, which
+  is currently a **read-only console surface** (`region tree`) and does **not**
+  drive emit.
+
+## Hypothesis for the kuna change (why it is LARGE)
+
+To reproduce angr's output you must **invert the guard conditions** so the
+exceptional path nests and the shared BODY becomes the fall-through join, then
+re-parent BODY out of the first `if`-then. This is **condition-based block-graph
+restructuring**, not a value-level Action/Rule:
+
+- It cannot be done as a single gated early-return in `blockaction.rs` — that file
+  can only *suppress* work, it cannot *synthesize* a new guard inversion and
+  re-parent a join.
+- The canonical small template (`kuna_loweredswitch.rs`) is an **S2 pcode-level**
+  Action and is not applicable to block-graph restructuring.
+- It is the **same root** as the known `gotoreduce` gap
+  (morton/my_message_callback, proposal **PR #54**) and the `irreducibleloops`
+  family (**PR #46**): all need kuna's read-only `RegionIdentifier` promoted into a
+  real **emit-path SAILR/Phoenix condition-based structurer**. That is a new pass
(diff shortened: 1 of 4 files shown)


mahaloz:
## Re-triage against today's `main` — confirmed still open, zero movement; implementation dispatched

Keeping this open. Re-measured on today's `main`: this is the one proposal in the batch where **nothing has improved**.

|---|---|---|---|
| gotos | 1 | **1** | 0 |
| labels | 1 | **1** | 0 |
| loops | 6 | 6 | 6 |
| ternaries | 1 | 1 | 1 |
| ifs | 16 | 16 | 20 |
| loc | 141 | 140 | 205 |

`kuna decompile .../x86_64/decompiler/newbury main` runs in 6.7 s and emits the offending construct byte-for-byte as described here: `if (v2 == getgid()) goto label_f6ec;` inside the SUID-guard chain, jumping forward into the shared BODY at `label_f6ec:` (setlocale / tzset / server loop).

The only visible change since 2026-06-25 is cosmetic: `branchflip` (#69, default-on) now emits the outer guard positively — `if (getuid() != 0) { …exceptional… } else { label_f6ec: BODY }` instead of `if (v2 == 0) { label_f6ec: BODY } else { … }`. The arms are swapped; the goto and label are unchanged.

### Verified negative results — this recommendation is now confirmed by measurement, not just by analysis

This PR recommended folding into the funded goto-reduction family rather than dispatching a standalone worker. That family has since landed in full — and **none of it touches this function**. Swept individually and in combination:

`gotoreduce` · `regionstructure` · `regionlooprefine` · `regionedgeorder` (default-off SAILR H2 edge ordering, tried **on**) · `ifelseflatten` · `crossjumprevert` · `taildup` · `earlyreturn` · `returndup` · `branchflip` · `iteregion` · `--mode aggressive`

All leave 1 goto + 1 label. So the sibling proposals it pointed at (#54 tail duplication, #46 irreducible loops) genuinely did **not** subsume this case — the transform needed here is a different one.

### Everything except the goto already matches or beats angr

6 loops, 1 ternary, 0 switches all match. kuna emits 16 ifs vs angr's 20 and 140 loc vs angr's 205 — kuna is more compact, and the ifs delta is angr's short-circuit expansion, not a defect.

### What changed since 2026-06-25: this is no longer "a new pass type"

This PR was ruled LARGE because the transform *"is a new pass type mutating the BlockGraph, not a value-level Action/Rule"*, and step 1 of its plan was *"promote `s7_regions/kuna_regionid.rs` from a read-only console surface into an emit-driving structurer."*

**That prerequisite is done.** `p8_structure/region_structurer.rs` is the emit-driving Phoenix/SAILR structurer today, default-on since #85 (DIV-12), with acyclic + ITE + cyclic + switch schemas and the SAILR virtualize-to-goto fallback. The remaining work is a new bounded schema on an existing mutable region tree.

### The transform

Recognize a multi-predecessor join `J` reached by (a) one arm of an `if` and (b) a forward goto from the sibling arm, where the sibling's non-goto remainder **terminates** (`fputs(...); return -1`). Invert the sibling's guard chain by de Morgan so the terminating path nests, hoist `J` to be the fall-through join after the merged guard, drop the goto + label.

The risk this PR flagged stands and is the whole game: a wrong guard inversion emits *wrong* C, not merely ugly C. Every predicate in this chain is a **call** (`getuid()`, `geteuid()`, `getegid()`, `getgid()`), so evaluation-order preservation is a hard constraint, not an assumption. The worker is instructed to decline any case it cannot prove.

An implementation worker is running on branch `feat/condjoin` (default-off option, two-pass stage test, four gates, speed + decbench measurement, hard stop on invalid C).

mahaloz:
## Implementation is up: #196

This proposal and its sibling turned out to share **one** root cause — Ghidra's `BlockBasic::isComplex` >2-statement readability budget gating the short-circuit fold in `CollapseStructure::rule_block_or` (and its verbatim twin `try_block_or` in the region structurer). Proposal #39 (`tr::build_spec_list` crossing edge) and proposal #56 (`newbury::main` SUID-guard goto) are two symptoms of that single gate.

They were implemented separately as #193 and #194, which collided (same struct field, same builder method, same gate line, same ElementId). Both are now closed and consolidated into **#196** — one option `condfold off|on|wide`, default-off, ElementId 4113, whose admission predicate is the union of the two rules.

Both criteria are met and measured against `main` @ `d09f21ce`:

- **#56** — `newbury::main` 1 goto / 1 label → **0 / 0** at `on`, matching angr.
- **#39** — the crossing `goto label_40242e` is gone at `on`; angr's exact goto count (4) lands at `wide`.

Notably, both proposals' **LARGE / "new pass type" verdicts were wrong**, and for the same reason: they were filed on 2026-06-25, before `regionstructure` (#85, DIV-12) made the Phoenix/SAILR region structurer the default emit-driving path. Proposal #56's step 1 — *"promote `kuna_regionid.rs` from a read-only console surface into an emit-driving structurer"* — was already done. Neither needed a new pass type; each needed one gate relaxation. Nor was a de Morgan rewrite or a purity analysis required: `negate_condition_rec` already distributes the NOT and **never swaps operands**, so #56's worry about reordering the `getuid`/`geteuid`/`getegid`/`getgid` predicates does not arise.

Leaving this open until #196 merges.

PR #65, opened 2026-06-25
Multi-format loader PR-9: MSVC C++ demangler arm

mahaloz:
## What

Adds the **MSVC C++ demangler arm** to `s1_demangle` (the `GnuDemanglerAnalyzer` analog) so MSVC-mangled names (the `cl.exe` `?`-scheme, e.g. `?foo@Bar@@QEAAHH@Z`) on PE/COFF render readable. Itanium (`cpp_demangle`: ELF/Mach-O C++ + MinGW-PE) and Rust (`rustc_demangle`) were already covered; MSVC was the gap (design `docs/multiformat-loader-design.md` §5.5, §8 PR-9).

## Before → after

A real MSVC-ABI COFF object (`clang -target x86_64-pc-windows-msvc` — emits the same `?`-scheme as `cl.exe`, which is unavailable on Linux):

| raw (`cl.exe` mangled) | name-only (`demangle_name`) |
|---|---|
| `?foo@Bar@@QEAAHH@Z` | `Bar::foo` |
| `?g@ns@@YAHHH@Z` | `ns::g` |
| `?freefunc@@YAHH@Z` | `freefunc` |

`kuna decompile msvc_mangled.obj freefunc --experimental-formats`:
```c
int4 freefunc(int4 a0)
{
  return a0 + 0x2a;
}
```
The function resolves **by its demangled name** (from `?freefunc@@YAHH@Z`), not the raw `?`-symbol.

## Implementation

- **`s1_demangle/mod.rs`**: `skip()` accepts a leading `?` **before** the `@`-version heuristic (MSVC embeds `@` *structurally* as namespace/type separators, so the version heuristic would otherwise reject every MSVC symbol). `demangle_raw`/`demangle_name` gain a `?`-gated MSVC arm (`NAME_ONLY` → qualified-name-only for the `::`-scope splitter; `COMPLETE` → full form for raw).
- **`elf_plt::strip_version`**: guarded to NOT truncate a leading-`?` name at `@`. It was cutting `?foo@Bar@@...` to `?foo` *before* the demangler saw it. No glibc-versioned symbol starts with `?`, so the ELF path is unchanged. (This was the load-bearing loader fix — caught by the loader-level test.)

## The e2e: REAL, not unit-only

`cl.exe` is unavailable on Linux, but `clang -target x86_64-pc-windows-msvc` produces a **genuine** MSVC-ABI COFF object with `?`-mangled symbols (verified `objdump -t`), so `verify_msvc_demangle.rs` loads + decompiles a real fixture (`msvc_mangled.obj`, 538 B), **not** a hand-faked symtab. It **RAN** (not skipped — `.sla` built), in 0.15s (release). The namespaced demangle (`Bar::foo`/`ns::g`, which nest into scopes) is pinned at the loader level + by the `s1_demangle` unit tests (the merge gate, no toolchain needed).

## Dependency / Cargo.lock

`msvc-demangler` 0.11.0 (MIT) added to the workspace + `kuna-analysis` manifests; `Cargo.lock` updated (pulls `bitflags` + `itoa`). Same dependency-substitution LOSS as the Itanium/Rust demangler deps and LOSS-005 (BFD→`object`) — recorded as **LOSS-250** in `docs/rust-port/losses.md`.

## Gates

- `make test` — **675/675 PARITY OK**
- `make test-stages` — **159/159 PARITY OK**
- `make rust-test` — **green** (exit 0; incl. the e2e + 8 new MSVC unit/loader tests: kuna-analysis lib 178 passed / 0 failed)

Default-off / additive (MSVC `?`-names only appear when present, and PE/COFF only load under `--experimental-formats`) ⇒ ELF byte-identical.

https://claude.ai/code/session_01FvRnfTgYKoZvS2TEvdYpWb


mahaloz:
APPROVE

PR #155, opened 2026-07-08
[PROPOSAL] kuna→IDA parity roadmap — 6 remaining feature proposals

mahaloz:
Draft/proposal PR (per the pipeline's `[PROPOSAL]` go/no-go convention) documenting the remaining `fmt/main` gaps against **IDA Pro 9.2**, after the landed parity work.

## Landed (already merged to `main`)
| PR | What |
|---|---|
| #149 | render fix — no stray `EAX`/`Unique` (register↔global merged highs render as the global) |
| #151 | name DWARF data globals (`dat_215120` → `max_width`, `goal_width`, `prefix*`) |
| #152 | `opflags_for` `INT_EQUAL` regression (from #150) |
| #153 | no-return: conditional jump to `.cold abort` still returns (`quotearg_style` error path restored) |
| #154 | reconcile #149 test witness with #151 naming |

## Proposals (this PR — `docs/features/<slug>/proposal.md` + a roadmap index)
| Proposal | Gap | Effort | Value |
|---|---|---|---|
| `ida-libc-extern-globals` | `dat_20a098` → `optind`/`stdin`/`stdout`/`optarg` (`.dynsym` `STT_OBJECT`) | Small–Med | High |
| `ida-dwarf-prototypes` | `undefined16 main(uint4,void*)` → `int main(int argc, char **argv)` | Med | High |
| `ida-string-ref-recovery` | `bindtextdomain(0x68d8,…)` → `"coreutils"` (mid-string char* ptr) | Med | Med |
| `ida-dwarf-enums` | `quotearg_style(4,…)` → `quoting_style::…` | Med | Low–Med |
| `ida-return-storage-inference` | `undefined16` return + `char[16]` phantom → `int` | **Large** | High |
| `ida-inlined-libc-idioms` | inlined `cmpsb`/`df` loop → `strcmp(v19,"-")` | **Large (multi-week)** | Med–High |

Each proposal carries the **confirmed root cause** (traced this session), the IDA/Ghidra reference, default-on policy, speed/risk, a testing plan, and an effort estimate. Suggested order + "where kuna already beats IDA" (char literals) are in `docs/features/README-ida-parity.md`.

The two `Large` items are flagged for explicit go/no-go and recommend approving a scoped first step (return-storage: a step-1 investigation; inlined-idioms: Phase 0 `df` folding) before committing to the full implementation.

Docs-only; no code changes.

https://claude.ai/code/session_01JJimTmF4YYsbdAumrkyPoL

### DIFF
--- a/docs/features/README-ida-parity.md
+++ b/docs/features/README-ida-parity.md
@@ -1,0 +1,48 @@
+# kuna → IDA Pro parity roadmap
+
+Tracks kuna's decompiled-output gaps against **IDA Pro 9.2** (reference oracle:
+`tests/x86_64/decompiler/fmt`, function `main`; captured via the `decompiler`/idalib CLI).
+Started from a `fmt/main` diff after the register↔global render fix.
+
+## Landed (merged to `main`)
+
+| PR | What | IDA gap closed |
+|---|---|---|
+| #149 | Render register/`unique` members of a global-merged high as the global (no stray `EAX`/`Unique<hex>`) | raw-register variables |
+| #151 | Name the program's DWARF data globals (`dat_215120` → `max_width`, `goal_width`, `prefix*`) | unnamed globals (DWARF) |
+| #152 | `opflags_for` `INT_EQUAL` regression (from #150) — was erroring functions in `decompile-all` | (regression) |
+| #153 | No-return: a conditional jump to a `.cold abort` still returns via fall-through (`quotearg_style` error path restored) | over-aggressive no-return |
+| #154 | Reconcile the #149 regmerge test witness with #151's naming | (test) |
+
+## Proposed (this directory — go/no-go)
+
+Ranked by value × tractability. Default policy for all: **default-on, no flag** (clearly-correct
+matches to IDA), except where a bring-up `--option` gate is flagged.
+
+| Proposal | Gap | Effort | Value | Notes |
+|---|---|---|---|---|
+| [`ida-libc-extern-globals`](ida-libc-extern-globals/proposal.md) | `dat_20a098` → `optind`/`stdin`/`stdout`/`optarg` | **Small–Med** | High | `.dynsym`/`.symtab` `STT_OBJECT` reader; reuses #151's install path |
+| [`ida-dwarf-prototypes`](ida-dwarf-prototypes/proposal.md) | `undefined16 main(uint4,void*)` → `int main(int argc, char **argv)` | **Med** | High | applies DWARF `DW_TAG_subprogram` return+params; also fixes #2 for DWARF binaries |
+| [`ida-string-ref-recovery`](ida-string-ref-recovery/proposal.md) | `bindtextdomain(0x68d8,…)` → `"coreutils"` | **Med** | Med | `char*` typing for **mid-string** readonly-char pointers |
+| [`ida-dwarf-enums`](ida-dwarf-enums/proposal.md) | `quotearg_style(4,…)` → `quoting_style::…` | **Med** | Low–Med | DWARF enum types → existing `push_enum_constant` render; best after prototypes |
+| [`ida-return-storage-inference`](ida-return-storage-inference/proposal.md) | `undefined16` return + `char[16]` phantom → `int` | **Large** | High | general/stripped return-storage recovery (core dataflow); high blast radius |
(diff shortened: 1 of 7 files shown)


mahaloz:
Closing: all six proposals are implemented. The roadmap docs under `docs/features/` are deliberately **not** merged — the behavior they describe now lives in `docs/spec/` (the live spec) and `docs/history.md` (the DIV registry), which is where it belongs.

| Proposal | PR | `fmt/main` before → after |
|---|---|---|
| `ida-libc-extern-globals` | #187 | `dat_20a098` → `optind` / `stdin` / `stdout` / `optarg` |
| `ida-dwarf-prototypes` | #188 | `undefined16 main(uint4,void*)` → `int main(int argc, char **argv)` |
| `ida-string-ref-recovery` | #189 | `bindtextdomain(0x68d8,…)` → `bindtextdomain("coreutils",…)`, `0x6f0c` → `"%s"` |
| `ida-dwarf-enums` | #190 | `quotearg_style(4,…)` → `quotearg_style(shell_escape_always_quoting_style,…)` |
| `ida-return-storage-inference` | #188 + #191 | the `char[16]` return whose high half read an uninitialized stack slot, gone |
| `ida-inlined-libc-idioms` (Phase 0) | #192 | `&v16[(uint8)v21 * -2 + 1]` → `&v12[1]`; the `df` variable gone |

The PRs are stacked in that order (#187 ← #188 ← #189 ← #190 ← #191 ← #192); merge them in sequence. Every one is green on all four gates with **no baseline re-pin** — 675/675 datatests, 261/261 stage tests, the workspace suite, and `check-spec`.

## No `--option` on any of them

Each turned out to be a case of kuna discarding evidence it had already recovered, which is the "strict correction, no flag" category rather than the "judgment call, gate it" one. Recorded as DIV-26 through DIV-31.

## Where the diagnosis differed from the proposal

Worth reading before reviewing, because in two cases the proposed fix would have been aimed at the wrong thing:

- **Return storage was not a separate feature.** The proposal treated `undefined16 main(...)` and the wrong signature as two items, one of them Large. They are one bug: the DWARF pass already builds a complete `PrototypePieces` for every defined subprogram and parks it on the function's symbol — but that snapshot was only ever read by a *caller*. Applying it to the function itself (#188) locks the `int` return, which never enters the recovery machinery that was building the bogus `RAX:RDX` join. #191 then handles the genuinely prototype-less case.

- **The direction flag was never unknown.** The proposal proposed proving `df == 0`. The processor spec already pins it and `ActionConstbase` materializes it at entry; what destroyed the constant was that the gcc *compiler* spec omits `DF` from `<unaffected>`, so every call planted `DF = INDIRECT(DF, call)`. The Microsoft prototype in the same file already lists it, and both x86 ABIs require the flag clear at function boundaries — so #192 states a documented platform guarantee instead of inferring one.

- **The mid-string gap was not in the string machinery.** The string manager reads happily from any address, and `ActionConstantPtr` already relaxes its exact-hit rule for character arrays. The reference it builds for an *interior* hit is a spacebase `PTRSUB` plus an `INT_ADD` of the residual, and constant folding collapses that pair back to the bare constant, discarding the type. An exact-start hit survives the identical path only because its residual is zero.

## Two things reviewers should look at closely

- **#191 rewrites two SPARC stage testcases.** `gh6990-returnpair.xml` asserted the *old bug* as expected default behavior; the repair fixes GH-6990 by default, on a different architecture with a different trigger than the x86 case it was written for, so the test now asserts the fix. `gh6882-sparcstructret.xml` used `return CONCAT44` as a proxy for "the tail was recovered" — the tail still is, the spurious join is not, so its assertion now names the real recovered expression. Both files keep their assertion counts; the stages-baseline diff is three renamed keys, not an absorbed regression.

- **#191's blast radius is measured, not asserted.** Whole-binary A/B on `fmt`: byte-identical with DWARF (locked prototypes never enter recovery), and with DWARF disabled, exactly **one** function signature changes out of the entire binary.

## Not done: Phase 1 of the idiom work

Re-rolling the now-legible `repz cmpsb` loop into a synthesized `memcmp`/`strcmp` call. That needs a loop-shape matcher with a provable-equivalence obligation plus a simplification of the trailing flag algebra — the proposal's own multi-week Phase 1, which it asks to gate behind a go/no-go after Phase 0. #192 records concretely what it would take, now that the loop is readable enough to pattern-match.

PR #188, opened 2026-07-29
feat(p4/protos): apply a function's own DWARF prototype to its own decompile

mahaloz:
Implements the `ida-dwarf-prototypes` proposal from #155 — and, as a consequence, the `fmt` case of `ida-return-storage-inference`.

> **Stacked on #187** (`feat/elf-data-symbols`). Base is that branch; merge #187 first.

## The gap

```c
// kuna (before)                      // IDA Pro
undefined16 main(uint4 a0,void *a1)   int __fastcall main(int argc, const char **argv, ...)
  char v4 [16];
  v4[0] = v16 ^ 1;   // real result     return (unsigned __int8)v16 ^ 1;
  v4[8] = v22;       // GARBAGE, uninitialized stack slot
  return v4;
```

Both symptoms are **one bug**. The DWARF pass already builds a complete `PrototypePieces` for every defined `DW_TAG_subprogram` — return type, parameter types, parameter names — and `commit_analysis_output` parks it on the function's global `FunctionSymbol`. That snapshot was only ever read by a **caller**: `ActionDefaultParams` copies a callee's pieces into the call site (which is why `fmt(FILE*, char const*)` typed its arguments correctly all along), while the function's own decompile ignored the signature and re-derived everything from data flow. The recovered prototype existed and was thrown away.

## What this does

The drive resolves the prototype from two sources, in precedence order:

1. an operator declaration for this run (`parse line extern …`) — unchanged, still wins;
2. the pieces parked on the function's own entry address (DWARF, or the library-prototype table for a named libc function).

One edit in `decompile_func_full_with_override_dyn` covers every consumer — the console, `decompile-all`, `decompile-project`, the Ghidra front-end, and the WASM build all drive through it.

### Why this also kills the bogus wide return

Locking the output is what collapses it. Unlocked, `ActionPrototypeTypes` starts return recovery, `Heritage::guard_returns` registers a trial for every output register the model characterizes (x86-64 SysV: `RAX` **and** `RDX`), and the cspec's `join_dual_class` output rule accepts the consecutive pair as one 16-byte return — materialized as `PIECE(RDX,RAX)`, typed `undefined16`, with the never-written `RDX` half picking up whatever the stack held. A known `int` return never enters that machinery. `fmt/main` now ends `return v13 ^ 1;`.

This closes the DWARF-having half of the return-storage proposal. The stripped/no-DWARF half is separate work.

## Policy

No `--option`. The prototype was recovered and then dropped on the floor for the function itself; applying it is a strict correction, not a judgment call. `apply_locked_prototype` already degrades gracefully — drops the prototype and decompiles unlocked — if storage assignment hits an unported seam. Recorded as DIV-27.

## Verification

- `make rust-test` — green (+3 new)
- `make check-spec` — OK
- Speed: 0.54s vs 0.54s on `fmt/main` — unchanged
- Testcase: `verify_dwarf_prototypes.rs` over `dwarf_globals_x86_64` (`bump(int4 n)`, `main(int4 argc,char **argv)`) and `regglobal_fmt_x86_64` (no `undefined16`, no `[16]` phantom, `return v13 ^ 1;`)
- Spec: `docs/spec/04-calls-and-prototypes.md` §4.2 (new "Seeding" subsection), `docs/spec/01-program-prep.md` §1.4

### DIFF
--- a/decompiler/crates/kuna-console/tests/verify_dwarf_prototypes.rs
+++ b/decompiler/crates/kuna-console/tests/verify_dwarf_prototypes.rs
@@ -1,0 +1,129 @@
+//! End-to-end gate for **applying a function's own DWARF prototype**: a `-g`
+//! binary's function renders its source signature — real return type, real
+//! parameter types, real parameter names — instead of a data-flow-derived
+//! approximation.
+//!
+//! The gap: the DWARF pass already built a full `PrototypePieces` for every
+//! defined `DW_TAG_subprogram` and parked it on the global `FunctionSymbol`, but
+//! that snapshot was read only by a *caller* (`ActionDefaultParams` copies a
+//! callee's prototype into the call site). A function's own decompile ignored it
+//! and re-derived everything from data flow, so `main` — described by
+//! `.debug_info` as `int main(int argc, char **argv)` — rendered
+//! `undefined16 main(uint4 a0, void *a1)`.
+//!
+//! The wide bogus return is the same bug's second symptom. With no locked output,
+//! return recovery registers a trial per output register the model characterizes
+//! (x86-64 gcc: RAX *and* RDX) and the cspec's `join_dual_class` output rule
+//! accepts the pair as one 16-byte return — emitted as a `char[16]` whose high
+//! half is an uninitialized stack slot. A known `int` return skips that entirely.
+//!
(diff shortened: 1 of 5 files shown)


mahaloz:

Its commit `85e57f0a` was never lost — it stayed on the stack tip `feat/df-unaffected` and

PR #189, opened 2026-07-29
feat(p9/strings): recover string literals through interior char-array pointers

mahaloz:
Implements the `ida-string-ref-recovery` proposal from #155.

> **Stacked on #188** (`feat/dwarf-prototypes`) → #187. Merge in order.

## The gap

```c
// kuna (before)                                  // IDA Pro
bindtextdomain(0x68d8,"/usr/local/share/locale")  bindtextdomain("coreutils", "/usr/local/share/locale")
textdomain(0x68d8)                                textdomain("coreutils")
error(1,*v30,0x6f0c,v29)                          error(1, *v30, "%s", v29)
```

Those addresses are not string *starts*. A compiler that merges string constants (`-fmerge-constants`, on at `-O2`) stores a literal that is a **suffix** of another literal only once: `"coreutils"` is bytes 4.. of `"GNU coreutils"` at `0x68d4`, and `"%s"` is bytes 12.. of `"%s: %s"` at `0x6f00`.

## Root cause

The gap was never in the string machinery — the string manager reads happily from any address, and `push_ptr_char_constant_ir` re-proves read-only-ness and scans to the NUL itself. The only question is whether the constant arrives carrying a character-pointer type.

`ActionConstantPtr` recognizes the interior pointer correctly (upstream deliberately relaxes its exact-hit requirement for character arrays, and kuna ports that relaxation), so `is_pointer` resolves `0x68d8` to the `char[14]` symbol at `0x68d4`. But the reference it builds for an **interior** hit is `PTRSUB(spacebase, 0x68d4)` plus an `INT_ADD` of the residual `4`, and constant folding collapses that pair straight back to the bare `0x68d8` — discarding the type with it. The constant then reaches the printer as `undefined8` and never enters the char-pointer arm.

An exact-start hit survives the *identical* path only because its residual is zero. That accident is the entire difference between the working and broken cases.

## What this does

Upstream repairs this downstream — `RulePtrsubCharConstant` rewrites the reference into a typed constant — but that rule is an un-ported stub here. So type the constant directly at the point where the evidence already exists: the covering symbol is a character-printable array, which is stronger proof than inspecting the bytes. The exact-hit path is untouched.

If the address turns out not to be read-only, or the bytes do not decode, the printer falls back to the integer render exactly as before — the worst case is today's output.

## Policy

No `--option`. The constant had already been *proved* to be a char pointer and the proof was being dropped. Recorded as DIV-28.

## Verification

- `make rust-test` — green (+2 new)
- `make check-spec` — OK
- Speed: 0.38–0.54s vs 0.54s on `fmt/main` — unchanged (one fewer `PTRSUB` built)
- Testcase: `verify_midstring_literals.rs` over a new purpose-built `midstring_x86_64` fixture — exact hit, interior hit, **and a negative control**: an interior pointer into a read-only `int` array must stay numeric — plus `regglobal_fmt_x86_64`
- Spec: `docs/spec/09-emission.md` §9.4


mahaloz:
### This really did merge — but into an intermediate branch, not `main`

Flagging so the "Merged" badge is not read as "shipped on `main`". This PR was part of the
six-PR IDA-parity stack (#187 → #188 → #189 → #190 → #191 → #192), and its base was
`feat/dwarf-prototypes` — the previous PR in the stack, not `main`. So it merged into that stack branch
and never reached `main` on its own. Bottom-up merging of the rest of the stack then deleted
each head branch, which was the *base* of the next PR, and GitHub auto-closed those orphaned
children (#188, #190, #192). (Note for future stacks: this repo has **`delete_branch_on_merge`
enabled repo-wide**, so the head branch is removed on every merge whether or not
`--delete-branch` is passed. Retarget the children onto `main` before merging, or merge the
tip only.)

Nothing was lost. This PR's commit `8a4275c9` stayed on the stack tip `feat/df-unaffected`
and has now reached `main` via the recovery PR **#195**, merge commit `d09f21ce`, together
`main` before merging: `make test` **675/675 PARITY OK**, `make test-stages` **261/261 PARITY
OK**, `make rust-test` green (4304 passed / 0 failed / 291 suites), `make check-spec` OK,
`kuna catalog --check` OK, and `docs/baseline.json` untouched (no re-pin). #187 had already
landed correctly as `ae4c5463`.

PR #190, opened 2026-07-29
feat(p5/types): recover DWARF enums so enum values render by member name

mahaloz:
Implements the `ida-dwarf-enums` proposal from #155.

> **Stacked on #189** (`feat/midstring-literals`) → #188 → #187. Merge in order.

## The gap

```c
// kuna (before)                     // IDA Pro
quotearg_style(4,(char *)file)       quotearg_style(quoting_style::shell_escape_always_quoting_style, (char *)file)
```

The decompiler could always render an enum constant by name — the printer's enum arm and its OR-of-flags decomposition are ported and tested. What was missing was the **type**: the DWARF pass flattened every `DW_TAG_enumeration_type` to its underlying integer and discarded the `DW_TAG_enumerator` children, so no varnode ever carried an enum type for the printer to act on.

## What this does

Builds a real enum `Datatype` — name, declared width, signedness, value→name map — and the existing machinery does the rest:

```c
char * quotearg_style(quoting_style s,char *arg)          // callee signature
v6 = quotearg_style(shell_escape_always_quoting_style,…)  // call site
if (m != mode_append) { … if (l == level_error) … }       // comparisons
```

Three details this needed:

1. **Width.** `get_type_enum` builds at the type factory's configured enum size, which `setup_sizes` derives from the architecture default — **8 bytes** on x86-64. A C enum is normally int-sized, and an 8-byte type will not bind to the 4-byte constant it describes. Added `get_type_enum_sized`, taking the declared width and signedness; upstream expresses the same thing through the `<type metatype="enum_uint" size="N">` decode path.
2. **Value masking.** Members are keyed by the value masked to the enum's width, so a negative member of a signed enum (`level_error = -1` in a 4-byte enum) is keyed `0xffffffff` — the value the constant Varnode actually carries.
3. **Look before building.** Filling the value map is a re-intern, and a real program repeats the same enum definition in every compilation unit that includes its header. Constructing it a second time is *rejected* as an altered definition, not treated as a no-op — which is why the first attempt silently fell back to `int` (4 of the 5 `quoting_style` DIEs errored out). The pass now resolves an already-recovered enum by name first.

Anonymous, memberless, or wrong-width enums fall back to the plain underlying integer — exactly what this arm always produced.

## Policy

No `--option`. This is DWARF the pass already reads and was throwing away. Recorded as DIV-29.

## Verification

- `make rust-test` — green (+3 new)
- `make check-spec` — OK
- Speed: load-time DWARF walk only. On a noisy host, 10 runs of `fmt/main` gave min 0.52s after vs 0.48s before (medians 0.60s vs 0.94s) — within noise
- Testcase: `verify_dwarf_enums.rs` over a new `dwarf_enums_x86_64` fixture (unsigned enum, signed enum with a negative member, as parameter types *and* as call arguments) plus `regglobal_fmt_x86_64`, which also covers the repeated-definition reuse path
- Spec: `docs/spec/05-types.md` §5.1, `docs/spec/01-program-prep.md` §1.4


mahaloz:

Its commit `41d1b141` was never lost — it stayed on the stack tip `feat/df-unaffected` and

PR #191, opened 2026-07-29
feat(p4/protos): drop the uncomputed half of a recovered return register pair

mahaloz:
Implements the general (no-DWARF) half of the `ida-return-storage-inference` proposal from #155. The DWARF-having half is #188.

> **Stacked on #190** (`feat/dwarf-enums`) → #189 → #188 → #187. Merge in order.

## The gap

With no prototype to lock the return, an x86-64 SysV function was modelled as returning a 16-byte `RAX:RDX` pair:

```c
undefined16 main(uint4 a0,void *a1)
  char v5 [16];
  v5[0] = v16 ^ 1;   // the real result
  v5[8] = v31;       // an uninitialized stack slot
  return v5;
```

That is not merely unreadable — it reads memory the function never wrote.

## Root cause

Return recovery registers one trial per output register the model characterizes (x86-64 SysV: `RAX` **and** `RDX`), and marks a trial active when its value survives ancestor realism. Ancestor realism asks whether a value could legitimately *reach* the RETURN — not whether the function meant to return it — so two shapes get through:

- a **callee-saved restore**, where the epilogue reloads a register from a frame slot the function only ever read (traced in the raw IR: `RDX = s0xffffffffffffffc8(i)`, an unwritten input);
- a **clobber at a synthesized return**, where the flow model turns a call that never returns into one and the registers hold the callee's INDIRECT creations (`RDX = [create] i0x1e02:376(free)`).

The spec's `join_dual_class` output rule then accepts the consecutive pair as one 16-byte return.

## What this does

A half carrying no value the function computed is not a return value. "Computed" is a bounded walk back through the operations that only *move* a value — copies, phis, indirects, piece/subpiece reshaping — stopping at the first one that produces one. An unwritten Varnode and an INDIRECT creation are uncomputed; a constant is computed (returning a literal is a real return); anything unclassifiable is computed, so an unfamiliar shape keeps today's answer.

**Timing is the load-bearing detail.** At recovery time the restore is still `COPY(LOAD(sp − k))`, indistinguishable from `return *p` — I confirmed this by instrumenting the walk, which is why the first attempt at this hook did nothing. The repair therefore runs in the one-shot tail, just before `ActionOutputPrototype` reads the storage and type off the RETURN, by which point heritage has resolved that load into a bare unwritten Varnode.

## Why a genuine wide return is safe

Two independent guards, both covered by tests:

1. Both halves of a real struct return are **computed** — built from constants, arithmetic, or loads through a pointer. A LOAD is not a move, so the walk stops there.
2. The rule only ever edits an existing *pair*; it never removes a lone recovered return value.

Where every half is uncomputed (the synthesized-return case) the low, first-in-class register is kept, so the function's output storage still agrees across every RETURN.

## Blast radius (measured)

- **DWARF `fmt`, whole binary**: byte-identical, 0 lines changed (prototypes are locked, so recovery never runs).
- **Stripped `fmt`, whole binary**: 201 lines changed out of 4283, and **exactly one function signature** — `main`, `undefined16` → `uint8`. Everything else is variable renumbering inside that same function.
- **675-assertion datatest corpus**: PARITY OK, no re-pin.

## It subsumes `returnpair` on GH-6990

The same repair fixes the SPARC32 `o0:o1` case that `option returnpair single` was written for — independent validation on another architecture with a different trigger (the save/restore window). `tests/stages/gh6990-returnpair.xml` asserted the *old bug* as expected default behavior, so it is rewritten to assert the fix; both of its passes now agree. `tests/stages/gh6882-sparcstructret.xml` used `return CONCAT44` as a proxy for "the tail was recovered" — the tail is still recovered, the spurious join is not, so the assertion now names the real recovered expression instead of a symptom of a second, unrelated bug. Both stage files keep their assertion counts (261/261); the baseline re-record is 3 renamed keys, not an absorbed regression.

`returnpair` stays as the blunt per-function instrument for a pair this rule judges genuine.

## Policy

No `--option`. The rule only ever removes output that provably carries nothing. Recorded as DIV-30.

## Verification

- `make rust-test` — green (+10 new: 7 unit, 3 e2e)
- `make check-spec` — OK
- Testcases: `kuna_returnuncomputed/tests.rs` pins the classification (constant, unwritten, copy-chase, arithmetic, **load**, phi-any, phi-none); `verify_return_uncomputed.rs` is the end-to-end witness plus **two negative controls** on a new `structreturn_x86_64` fixture — a struct pair built by arithmetic and one loaded through a pointer, both of which must keep their 16-byte return
- Spec: `docs/spec/04-calls-and-prototypes.md` §4.4


mahaloz:
### This really did merge — but into an intermediate branch, not `main`

Flagging so the "Merged" badge is not read as "shipped on `main`". This PR was part of the
six-PR IDA-parity stack (#187 → #188 → #189 → #190 → #191 → #192), and its base was
`feat/dwarf-enums` — the previous PR in the stack, not `main`. So it merged into that stack branch
and never reached `main` on its own. Bottom-up merging of the rest of the stack then deleted
each head branch, which was the *base* of the next PR, and GitHub auto-closed those orphaned
children (#188, #190, #192). (Note for future stacks: this repo has **`delete_branch_on_merge`
enabled repo-wide**, so the head branch is removed on every merge whether or not
`--delete-branch` is passed. Retarget the children onto `main` before merging, or merge the
tip only.)

Nothing was lost. This PR's commit `7e2c3748` stayed on the stack tip `feat/df-unaffected`
and has now reached `main` via the recovery PR **#195**, merge commit `d09f21ce`, together
`main` before merging: `make test` **675/675 PARITY OK**, `make test-stages` **261/261 PARITY
OK**, `make rust-test` green (4304 passed / 0 failed / 291 suites), `make check-spec` OK,
`kuna catalog --check` OK, and `docs/baseline.json` untouched (no re-pin). #187 had already
landed correctly as `ae4c5463`.

PR #192, opened 2026-07-29
feat(p4/protos): state the x86 direction-flag ABI guarantee (fold `* -2 + 1` strides)

mahaloz:
Implements **Phase 0** of the `ida-inlined-libc-idioms` proposal from #155 — the prerequisite the proposal recommends approving first, and independently the larger readability win of the two.

> **Stacked on #191** (`feat/return-uncomputed`) → #190 → #189 → #188 → #187. Merge in order.

## The gap

gcc inlines `strcmp(file, "-")` in `fmt/main` as a `repz cmpsb` (raw bytes at file offset `0x1c77`). Every x86 string instruction scales its pointer step by `1 − 2·DF` — forward when the direction flag is clear, backward when set — and SLEIGH lowers that faithfully, so the flag reached the output as a live variable and scaled every step:

```c
// before
uint1 v21;                            // df
v21 = 0;
...
v15 = &v16[(uint8)v21 * -2 + 1];
v13 = &v14[(uint8)v21 * -2 + 1];

// after
v2 = &v12[1];
v1 = &v11[1];
```

## Root cause (not what the proposal guessed)

The proposal assumed `DF` was genuinely unknown and proposed proving `df == 0`. It is not unknown. The **processor** spec already pins it — `<tracked_set><set name="DF" val="0"/>` — and `ActionConstbase` materializes `DF = 0:1` in the entry block precisely so the stride can fold.

What breaks the chain is the **call**. The gcc **compiler** spec's default prototype omits `DF` from `<unaffected>`, so SSA construction must assume a callee might change it and plants `DF = INDIRECT(DF, <call>)` at every call site. Constant propagation cannot see past that, so the fold never happens in any function that calls anything — which is every interesting one.

## Why asserting it is correct, not a heuristic

Both x86 ABIs require the direction flag to be clear at every function boundary: the System V psABI states it must be clear on entry and on return, and the Windows x64 ABI says the same. A callee that sets it must clear it before returning. The **Microsoft prototype in the very same compiler spec already lists `DF` as unaffected** — the gcc one simply omits it. So this states a rule the platform already guarantees rather than inventing one.

Applied conservatively:

- only where the spec is **silent** — a model that names `DF` either way has made a deliberate statement and is left alone;
- only on languages that **have** a `DF` register — the assertion is keyed on the SLEIGH register name, and a lookup miss is the exit, so every non-x86 target is a structural no-op.

No vendored spec file is modified.

## Policy

No `--option`. Recorded as DIV-31.

## Verification

- `make test` — 675/675, **PARITY OK** (no re-pin)
- `make rust-test` — green (+2 new)
- `make check-spec` — OK
- Testcase: `verify_direction_flag.rs` — `fmt/main` (which calls several functions *before* reaching the string op, so it needs the flag to survive a call) has no `* -2 + 1` stride and no `// df` variable, and still has its loop; plus an AArch64 fixture as the non-x86 no-op guard
- Spec: `docs/spec/04-calls-and-prototypes.md` §4.3
- Blast radius on the `fmt` binary: all **12** DF-scaled strides and all **4** `df` variables gone, 494 of 4283 lines changed (variable renumbering in the affected functions), and **zero** function signatures changed

## What Phase 1 still needs

With the stride folded, the loop reads as a recognizable byte comparison:

```c
v10 = 2;  v11 = file;  v12 = (uint1 *)0x6f52;
do {
  if (v10 == 0) break;
  v10 = v10 + -1;
  v2 = &v12[1];  v1 = &v11[1];
  v16 = *v11 < *v12;  v17 = *v11 == *v12;
  v11 = v1;  v12 = v2;
} while (v17);
if ((!v16 && !v17) != v16) { … }
```

IDA renders that as `if (!strcmp(v19, "-"))`. Getting there needs a loop-shape matcher (counted loop, two pointers stepping by one, byte compares feeding CF/ZF), a proof that the shape computes `memcmp(a, b, n)` semantics — note `repz cmpsb` is `memcmp`, not `strcmp`; gcc uses it here only because it knows the literal's length — a rewrite to a synthesized call, and separately a simplification of the trailing flag algebra `(!v16 && !v17) != v16`. That is a pattern-recognition pass with its own correctness obligation, in the same weight class as the SAILR structurer ports, and is the proposal's own multi-week Phase 1. It is **not** in this PR.


mahaloz:

Its commit `99a04679` was never lost — it stayed on the stack tip `feat/df-unaffected` and

PR #193, opened 2026-07-29
feat(p8): condfold — short-circuit condition folding across a non-trivial sibling block

mahaloz:
Implements proposal #39 (`tr-build-spec-list-ea32be`). Ships behind `option condfold`, **default-OFF**.

## The gap

Ghidra's `ruleBlockOr` declines the `A || B` short-circuit fold whenever the sibling condition block is *complex* (`BlockBasic::isComplex`, block.cc:2403 — more than two statements, where the trailing branch already counts as one). One spill or address computation parked in front of the second test is enough, and kuna then emits the second condition as its own block that branches back into the first arm's clause — a crossing `goto` where angr renders one compound condition. On the proposal's own target, coreutils `tr::build_spec_list`, that is literally:

```c
if (v2) {
label_40242e:
    ... shared clause ...
}
else {
    result_idx = &v10;                              // the prefix that declines the fold
    if (es_match(es,v12,'=')) goto label_40242e;    // the crossing edge
```

angr does not decline here: `_match_acyclic_short_circuit_conditions` (types a–d, phoenix.py:2578) checks `_is_single_statement_block` and, on a multi-statement sibling, wraps the operand in a `MultiStatementExpression` — a C **comma expression** — under the `MAX_ONE_CALL` policy (≤1 call, ≤5 non-label statements), refusing outright on a non-terminal `ConditionalJump` or any `Jump`.

## What this ships

A second, narrower disjunct at the single `is_complex(orblock)` decline, in **both** structurers. A complex sibling is accepted when it is a `BlockCopy` of ONE `BlockBasic` that `kuna_condfold.rs::condfold_eligible` approves; the prefix then renders inside the `&&`/`||` operand as a comma expression, which the printer already supports (`comma_separate`):

```c
if ((v2) || (result_idx = &v10, es_match(es,v12,'='))) {
```

**Three corrections to the proposal, all verified against the code.** (1) It classified this LARGE / "new pass type" / "needs C-printer support". kuna already has the `BlockCondition` node kind, the `&&`/`||` printer, the comma-expression rendering for a multi-statement operand, and *two* independent short-circuit schema implementations — the whole delta was one boolean at a decline. (2) angr does **not** hoist the prefix; it builds a comma expression in place. Hoisting p-code between `BlockBasic`s at P8 (post-SSA, MULTIEQUALs already placed) would be far riskier and is strictly unnecessary — **no p-code is moved, reordered, cloned, or deleted by this PR**. (3) On `build_spec_list` the region structurer does *not* converge (`[rs] STUCK: 22 components`, `ok=false`) and `ActionBlockStructure` falls through to `CollapseStructure`, so `blockaction.rs` is what emits the measured output; both paths are patched anyway since they are verbatim copies.

## Measurements

`gotos`/`labels`/`ifs`/`loops` with comment lines stripped. Two vehicles: `tr.o` directly, and the proposal's documented full-fidelity ELF (`gcc -no-pie -nostdlib -nostartfiles -static -Wl,-Ttext-segment=0x400000 tr.o stubs.c`, `build_spec_list` @ `0x402370`) — the vehicle on which the proposal's named defect is actually present.

| `build_spec_list` | loc | gotos | labels | ifs | loops | `\|\|` | `&&` |
|---|---|---|---|---|---|---|---|
| **angr 9.2.223 reference** | 123 | **4** | 3 | 15 | 2 | 1 | 1 |
| linked ELF — `condfold off` | 118 | 6 | 4 | 16 | 3 | 0 | 0 |
| linked ELF — `condfold on` | 115 | 5 | 3 | 15 | 3 | 1 | 0 |
| linked ELF — `condfold wide` | 109 | **3** | 2 | 14 | 3 | 1 | 1 |
| `tr.o` — `condfold off` | 112 | 5 | 3 | 15 | 3 | 1 | 0 |
| `tr.o` — `condfold on` | 112 | 5 | 3 | 15 | 3 | 1 | 0 |
| `tr.o` — `condfold wide` | 107 | **3** | 2 | 14 | 3 | 1 | 1 |

**Success criterion met.** On the linked ELF the crossing `goto label_40242e` is gone and the compound condition appears, at `on`. On `tr.o` the criterion asked for 5 → 4 (angr parity); `wide` gives 5 → 3. The pre-existing two-term fold `(es_match() != '\0') || (es_match() != '\0')` is present in all three settings — not regressed.

### Why the option is three-valued

`on` is exact angr parity (threshold 5, `MAX_ONE_CALL`). `wide` raises **only** the statement cap to 9. kuna's *printed* statement granularity is finer than angr's AIL for the same source block: on the `tr.o` third-term candidate angr renders `find_closing_delim(ptr, v11, *((char *)(v10+cur)), &v7)` as one AIL statement, while kuna recovers two arguments and prints the operand computations plus two address-tied copy shadows (`v14 = v1; v14 = v14;`) separately — 7 printed statements for the same block. The call cap and every soundness/rendering guard are identical at both levels.

### Bonus: PR #53's residual is also closed

The brief flagged `true_ubuntu1804::usage` as having a measured **+2 ifs** residual vs angr from this same root (angr short-circuits `if (!v30 || !strncmp(v30, "en_", 3))` twice; kuna emits two nested ifs). Measured:

| `usage` | loc | gotos | labels | ifs | `\|\|` | `&&` |
|---|---|---|---|---|---|---|
| off | 107 | 1 | 1 | 9 | 0 | 1 |
| on | 107 | 1 | 1 | 9 | 0 | 1 |
| **wide** | 103 | 2 | 2 | **7** | 1 | 2 |

`ifs 9 → 7` is exactly the residual, and both `!v30 || !strncmp(...)` pairs now fold into angr's shape. It costs +1 goto and −4 loc. Different binary, different function, same gate — good evidence of generality. (The recon agent could not confirm this statically; the empirical answer is that the block exceeds the angr-parity cap of 5 printed statements but not the wide cap of 9, which is why `on` is a no-op here.)

### Aggregate sanity sweep — invalid C is a hard stop

`kuna decompile-all --json` at off/on/wide over 5 real binaries (the default P8 goto-quality set ON — the real configuration, not ablated). Every function whose text *changed* vs off was syntax-checked by wrapping it in a permissive C translation unit and running `gcc -fsyntax-only`, filtering undeclared-identifier and unknown-type diagnostics but **not** syntax errors. The checker was validated to reject both failure modes this option could produce (a label inside an expression; an unbalanced condition).

| totals over 1580 functions | off | on | wide |
|---|---|---|---|
| gotos | 931 | 915 | **902** |
| ifs | 5077 | 5001 | **4961** |
| loc | 42963 | 42662 | **42492** |
| functions changed | — | 33 | 74 |
| **functions valid OFF but invalid ON** | — | **0** | **0** |

Per binary: `du` 212/212/204 gotos, `dd` 160/153/152, `basenc` 123/121/120, `mv_-O2` 374/370/366, `true_ubuntu1804` 62/59/60. Monotone improvement in every aggregate metric; the single per-binary inversion is `true_ubuntu1804` gotos at wide (59 → 60), which is the `usage` case above trading +1 goto for −2 ifs.

### Speed

The box was under heavy third-party load during measurement (load average 21–35 from another user's Ghidra headless jobs). `scripts.pipeline.timeit` times all OFF samples then all ON samples, so drift does not cancel: successive runs of the *same* measurement reported +7.9%, +24.1% and +72.0%, and the same configuration's median moved from 209 ms to 568 ms between runs. Those are machine noise.

The honest measurement is **interleaved** (alternate off/on/wide per iteration) on **CPU time** (user+sys of the child, insensitive to contention), 25 samples:

| | off | on | wide |
|---|---|---|---|
| CPU median (ms) | 323.66 | 322.97 | 322.79 |
| delta | — | **−0.21%** | **−0.27%** |

Interleaved wall-clock over 21 samples agrees in sign (−4.27% / −2.43%). Well inside the 5% budget, and structurally expected: one extra predicate inside the loop that already calls `bb_is_complex` (skipped entirely when off), plus one `BTreeSet` lookup at an existing decline. Raw `timeit` numbers are recorded in `record.json` and labelled unreliable.

## Soundness

The fold is **rendering-only**. `BlockGraph::new_block_condition` re-parents two existing structuring nodes. Soundness rests on a precondition `rule_block_or` already enforces and this option never weakens: `orblock.size_in() == 1`, so the sibling's statements cannot run on a path where the compound condition is not evaluated; C's short-circuit rule evaluates the right operand exactly when control used to fall into the sibling; and the comma operator is a sequence point, so the prefix still runs before the test. The semantic precondition set beyond `ruleBlockOr`'s own is therefore **empty** — every added guard is a rendering-validity or readability guard:

1. **`BlockCopy` of one `BlockBasic` only.** A `BlockList`/`BlockIf`/`BlockCondition` operand can render braces, multiple lines, or a label inside the parentheses — invalid C.
2. **`!is_interior_goto_target` is never relaxed** (pre-existing). `emit_block_copy` emits `emit_any_label_statement` *first*, so a labelled sibling would print `label_x:` inside an expression.
3. **Bounded prefix** (≤5 statements at `on` / ≤9 at `wide`, ≤1 *statement-root* call), counted with the printer's own skip rules (`not_printed` / markers / implied-out / bare `CPUI_BRANCH`) so the cap governs exactly what appears. See *The call cap counts statement-root calls only* below — the implied-out skip runs before the call test, so the call bound is looser than angr's `MAX_ONE_CALL`.
4. **No non-terminal branch op** — mirrors angr's `_build_multistatementexpr_statements`; a `goto` is not an expression.
5. **No instruction comment in the block.** `emit_basic_block_ops` suppresses `emit_comment_group` under `comma_separate`, so folding a block carrying a `WARNING:` line would *silently delete it*. Decline instead.
6. **The folded node is force-marked complex.** `BlockCondition::isComplex` delegates to sub-block 0, so a node whose *right* operand is a comma chain would look trivial to `rule_block_while_do` and the whole chain would be swallowed into a `while(...)` header. Unit-tested.

**`is_complex()` itself was not changed** — the relaxation is a disjunct at the two `rule_block_or`/`try_block_or` call sites only, so the other two callers (`rule_block_while_do` and its region-structurer twin) see the unchanged verdict. That is why the 675-assertion corpus is untouched.

## Gates

| gate | result |
|---|---|
| `make test` | **PARITY OK**, 675/675 |
| `make test-stages` | **PARITY OK**, 268/268 (was 261; +7 from `ghangr-condfold.xml`) |
| `make rust-test` | green (exit 0, 0 failed targets) |
| `make check-spec` | OK in lenient **and** strict mode |

`docs/baseline.json` was **not** re-pinned. `docs/baseline-stages.json` was re-recorded (a new stage test was added, which is the sanctioned re-record).

The stage test `tests/stages/ghangr-condfold.xml` is a self-contained bytechunk with the exact schema shape (no relocations, no calls): pass 1 (`option condfold off`) asserts the bug — two nested ifs and a crossing `goto label_401019` — and pass 2 asserts the fix — one compound condition carrying the sibling prefix as a comma expression, with the goto/label pair gone. Also 8 unit tests (5 hand-built-graph gate tests in `blockaction/tests.rs`, 3 in `kuna_condfold.rs`).

## What I did NOT close

- **Default-ON.** Needs its own 0/675 ablation and a DIV row. The aggregate sweep is supporting evidence, not the ablation.
- **The `region_structurer.rs` mirror is currently unexercised end-to-end.** On every function measured — both `build_spec_list` vehicles, `usage`, the stage witness, all 13 changed functions in `dd` and all 7 in `basenc` — the region structurer reports `ok=false` and falls through to `CollapseStructure`. I verified this is *pre-existing* (identical `[rs] STUCK` trace on `main`'s prebuilt binary) and that the region structurer does converge elsewhere (`basenc::isbase64`, `basenc::quote`). The mirror ships because the two functions are verbatim copies and shipping only one would silently no-op wherever the other engine converges — but I am not claiming end-to-end coverage for it.
- **Making the region structurer converge on `build_spec_list`** is a separate, larger opportunity.
- **angr's type_b / type_d (with-else) arms.** This matches angr only through Ghidra's `ruleBlockOr` topology; graphs those checks reject remain unfolded.
- **decbench `rescore`.** `scripts.decbench.mine --select 28` completed and regenerated the pool, but the search for a mined case whose output condfold changes did not finish inside the ~10 minute box on the loaded machine. No `ged_before`/`ged_after` is claimed. Not a blocker: the option is default-OFF, so the benchmark cannot move without an explicit opt-in.
- **The call cap counts statement-root calls only — an explicit open choice for the reviewer.** In `kuna_condfold.rs::condfold_eligible` the implied-output skip (`continue`) runs **before** the `if o.is_call()` test, so a call inlined into the sibling's own condition expression is never charged against `MAX_PREFIX_CALLS = 1`. The enforced bound is therefore "at most one call printed as its own comma-chain element", not "at most one call in the rendered operand". Measured: 1 of the 46 new call-bearing folded operands renders two calls — `mv_-O2 copy_internal` at `wide` renders `cached_umask(...)` (statement root, counted) alongside `fchmod(...)` (implied into the test, not counted). This is a **readability** bound, not a soundness bound: the fold moves no p-code and the sequencing argument above is independent of the call count, so the looser count cannot produce wrong C. Two ways forward, reviewer's call:
  1. **Accept statement-root-only counting as the documented semantics** (what this PR does — the module doc, the `MAX_PREFIX_CALLS` doc comment, the `phases.toml` row and `docs/spec/08-structuring.md` all now describe the actual behaviour and drop the "exact angr parity" claim for the call cap; `on` is still exact parity on the *statement* threshold).
  2. **Tighten it to charge implied calls too.** A two-line reorder (hoist the `is_call` test above the implied-output skip), but it *tightens* the gate, so the metric table, the aggregate sanity sweep and the stage-test witness would all have to be re-measured before it could ship.
- **The comment guard reads `Funcdata`'s buffered analysis comments.** A user comment already resident in the console `Architecture`'s comment database is not visible from `Funcdata` at P8. The analysis `WARNING:` lines this option could actually destroy are covered.

Full measurement record: `docs/features/condfold/record.json`.

Closes #39

https://claude.ai/code/session_015PVb6KdMKGV5rGEooSGHr4


mahaloz:
## Merge-order note — #193 (`condfold`) and #194 (`condjoin`) collide on the option count

These two PRs were developed in parallel in isolated worktrees, so neither saw the other. They are **independent** — different modules, different options, no shared logic — and **either merge order works**. But they touch the same hard-coded counters, so:

> **Whichever merges SECOND must rebase on `main` and bump every settable-option count assertion `77 → 78`, then re-record `docs/baseline-stages.json`.**

Both PRs currently move the count `76 → 77`. Merging the first is a no-op for the second's diff; merging the second without the bump fails `make rust-test` and `make test-stages` loudly (this is self-announcing, not silent), so nothing can slip through — this note just saves the rediscovery.

### Exact files the second merge must touch

| File | What to change |
|---|---|
| `decompiler/crates/kuna-decomp/src/p0_knowledge/kuna_phases/tests.rs` | `fn settable_count_is_77` → `_78`; `kuna_num_settables()` and `SETTABLE_TABLE.len()` assertions `77` → `78`; `fn tier_counts_are_16_core_37_transform_24_analysis` → `..._38_transform...` and its `(16, 37, 24)` → `(16, 38, 24)` (both options are `tier = "transform"`); `json.matches("},\n").count()` `76` → `77` |
| `decompiler/crates/kuna-decomp/tests/catalog_bytecompat.rs` | `fn fixture_has_all_77_settables` → `_78`; the `"option": ` / `"tier": ` / `"symptoms": ` count assertions `77` → `78` |
| `decompiler/crates/kuna-decomp/tests/fixtures/phase_catalog.json` | Regenerate — it is a byte-exact fixture, not a count. Procedure is in the `catalog_bytecompat.rs` module doc: `printf 'openfile write /tmp/cap.json\nphase catalog\nclosefile\nquit\n' \| decompiler/target/release/decomp_dbg -s "$PWD/specs"` then copy over the fixture |
| `tests/stages/kuna-catalog.xml` | KUNA-CATALOG **#6** (`"source_decompiler": "angr"`) `min/max="31"` → `32` and add the other option's name to the test name list; KUNA-CATALOG **#7** (`"change_kind": "structure-recovery"`) `min/max="20"` → `21` and likewise |
| `decompiler/crates/kuna-base/src/xml.rs` | The datatest corpus **file** count `assert_eq!(count, 165, "corpus file count drifted")` → `166` (both PRs add one `tests/stages/*.xml`), and extend the explanatory comment |
| `docs/options.md` | Generated — `decompiler/target/release/kuna catalog --markdown > docs/options.md` |
| `docs/baseline-stages.json` | Re-record: `kuna test --datatests --datatests-dir tests/stages --save-baseline docs/baseline-stages.json` (sanctioned — a stage test was added) |

**#194-only extra:** `kuna_phases/tests.rs` also renames `option_values_live_value_present_for_28_suppressed_for_39` → `_40`. `condfold` does not touch that test, so if #194 is the second merge it becomes `_41`; if #194 merges first, #193's rebase leaves it at `_40`.

`docs/baseline.json` (the 675-assertion datatest corpus) is **not** touched by either PR and must **not** be re-pinned by the rebase.

### ElementId — already deconflicted

Both branches originally allocated `ElementId` **4113** (`main`'s highest is 4112 = `switchreturn`). That is now resolved on the branches, so the reviewer merging either one does not need to act:

- `condfold` (#193) keeps **4113**
- `condjoin` (#194) moves to **4114**

The provenance comments on both branches now record the other's allocation, and #193's comment additionally corrects a factual error (it named 4112 as `loopbreak`; on `main` 4112 is `switchreturn` and `loopbreakrecovery` is 4092).

Nothing else in either diff overlaps.

mahaloz:
## Superseded by #196 — consolidated with #194 into one option

Closing in favour of **#196**, which carries this work forward in full. Nothing here is discarded.

### Why

This PR and #194 (`condjoin`, proposal #56) turned out to be **the same feature, discovered twice**. Both relax the identical gate — the `is_complex(orblock)` decline inside `CollapseStructure::rule_block_or` and its verbatim twin `try_block_or` in the region structurer — rooted in Ghidra's `BlockBasic::isComplex` >2-statement readability budget. They collided textually too: both added a `CollapseStructure` field named `condfold_blocks`, both added a builder named `with_condfold_blocks`, `kuna_condjoin.rs`'s entry point was even named `compute_condfold_blocks`, and both bumped the option count 76 → 77.

Two catalog entries relaxing one gate cuts against the option surface being the product, so they ship as one.

### What survived into #196

Everything. An independent reviewer diffed the consolidation against both originals:

- `prefix_eligible` (**Rule A**) is this PR's `condfold_eligible`, byte-identical apart from the rename and `MAX_PREFIX_CALLS` → `MAX_PREFIX_ROOT_CALLS`.
- The `condfolded` force-complex marking — this PR's contribution — is preserved verbatim, **and it closed a defect #194 had on its own**: without it, `is_complex(BlockCondition)`'s `getBlock(0)` delegation let a later fold skip #194's caps entirely.
- Unit tests went 6 → 9 (two new nested-`BlockCondition` tests plus a cap-enforcement test).
- `ghangr-condfold.xml` is retained, differing only in its comment block.
- ElementId **4113** is retained; `condjoin`'s 4114 was released.

### What changed on the way

The reviewer found that unioning the two predicates broke this PR's value surface. Rule B admits by statement *shape*, scored with Ghidra's `calc_explicit` approximation — which reported **2** on a block that prints **seven** statements. That meant `on` silently performed the folds this PR had correctly argued required `wide`, and the two levels came out byte-identical on every named target.

#196 fixes it by factoring this PR's printed-statement walk into a shared measure enforced by *both* rules against the level cap. The widest comma operand created by the option is now **5 at `on` / 7 at `wide`** (was 7/7), so `on` is genuinely the conservative setting again and your original finding — that the `tr` case needs the wider cap — is restored and measured.

Also corrected there: this PR's documented `MAX_PREFIX_CALLS ≤ 1` cap was never enforced for calls *implied* into the condition (witness: `mv -O2 copy_internal` at `wide` renders both `cached_umask(...)` and `fchmod(...)`). #196 keeps statement-root semantics deliberately and renames the constant `MAX_PREFIX_ROOT_CALLS` so the name is the contract.

All measurements were re-taken against `main` @ `d09f21ce` — the six-PR IDA-parity stack landed in between and changes emitted types, names, enums and string literals, so every number in this PR is stale.

Branch `feat/condfold` is left in place for reference.

PR #194, opened 2026-07-29
feat(p8): condjoin — condition-based join restructuring (guard inversion) to remove a forward goto into a shared body

mahaloz:
Implements proposal **#56** (`newburry-main-57ccb1`). Target: lighttpd `server.c::main`, angr corpus binary `newbury` (that on-disk spelling; the proposal title says "newburry") at `0x40f696`.

Ships as **`option condjoin on|off`, default-OFF**.

## The gap

kuna emitted **1 goto / 1 label** where angr emits **0 / 0** — numerically identical to the proposal-era column, i.e. zero improvement on this function since #56 was filed, despite the whole SAILR family landing in between.

```c
if (getuid() != 0) {
   v2 = geteuid();
   if (v2 == getuid()) {
     v2 = getegid();
     if (v2 == getgid()) goto label_f6ec;      // forward goto INTO the sibling arm
   }
   v2 = -1; fputs("Are you nuts ...", stderr);
} else {
label_f6ec:
   setlocale(2,"C"); tzset(); do { /* server loop */ } while (...);
}
```

## Root cause — and why the proposal's framing made it look larger than it is

The four blocks are a textbook `ruleBlockOr` cascade:

```
B0 @f69a: getuid();            test; je f6ec   -> {BODY, B1}
B1 @f6b0: geteuid(); getuid(); cmp; je f6dc    -> {B2, B3}
B2 @f6dc: getegid(); getgid(); cmp; jne f6c0   -> {B3, BODY}
B3 @f6c0: r = -1; fputs(...)   -> RET      (in-degree 2: B1, B2)
BODY @f6ec: the server loop    -> RET      (in-degree 2: B0, B2)
```

Every gate of `CollapseStructure::rule_block_or` holds for `(B1,B2)` and then for `(B0, B1||B2)` — 2-out, no goto-outs, non-switch, `orblock` single-in, not an interior goto target, no back edge, shared clause block — **except one**: `is_complex(orblock)`. `Funcdata::bb_is_complex` (Ghidra `BlockBasic::isComplex`) calls a block complex once its statement count exceeds 2, counting the branch as one and each call as one; B1/B2 hold two calls each, so both score 3. That decline is what leaves `BODY` with two predecessors and forces one in-edge to be virtualized.

The proposal classified this **LARGE** ("invert the guard chain by de Morgan and hoist the join") and warned that a wrong inversion emits *wrong* C because the predicates are calls that cannot be reordered. **Neither holds:**

* No hand-rolled inversion is needed. `negate_condition_rec` + `new_block_condition` already perform the de Morgan orientation and merge the join edge, and `ruleBlockIfElse` then hoists the join.
* No reordering is possible, so no purity analysis is needed (see below).

Also worth recording: the region structurer **bails** on this function (`[rs] STUCK: 25 components remain (round 88)`, then `STUCK: 7`), so the C actually comes from `CollapseStructure`, not `region_structurer.rs`. Confirmed independently by `--option regionstructure off` producing byte-identical output.

## What the option does

It **relaxes that one gate** — in both engines, since `region_structurer::try_block_or` carries the identical predicate. New module `p8_structure/kuna_condjoin.rs` (ELEM 4113) precomputes, over the live `bblocks` op lists, the set of complex blocks that are nonetheless safe to absorb; `CollapseStructure::condjoin_admits` / `RegionStructurer::condjoin_admits` consult it beside `is_complex`.

`isComplex` is a **readability budget, not a correctness gate**: `PrintC::emit_block_condition` already sets `modifiers::COMMA_SEPARATE` on a folded condition's right operand, and `emit_basic_block_ops` under that modifier walks *every* op and emits them comma-separated with the semicolon suppressed. Ghidra itself exercises that path for the one-extra-statement case its threshold admits. This is the same escape hatch angr has: when `_is_single_statement_block` is false, `phoenix._match_acyclic_short_circuit_conditions_type_{a,b,c,d}` wrap the operand in a `MultiStatementExpression` instead of declining (`MultiStmtExprMode`, `multistmtexpr_stmt_threshold`). kuna's `COMMA_SEPARATE` *is* `MultiStatementExpression`.

Result on the target:

```c
if ((getuid() != 0) && ((v2 = geteuid(), v2 != getuid() || (v2 = getegid(), v2 != getgid())))) {
  v2 = -1;
  fputs("Are you nuts ? Don't apply a SUID bit to this binary\n",dat_5f140);
} else {
  setlocale(2,"C"); tzset(); do { /* server loop */ } while (dat_5f278 != 0);
}
return v2;
```

## Soundness

**S1 — no reordering, and this is structural, not a check.** `new_block_condition` only *relabels* two blocks that are already sequenced `bl` -> `orblock` in the CFG as the left and right operands of a `&&`/`||`. C's `&&`/`||` short-circuit strictly left to right and the comma operator sequences strictly left to right, so the emitted expression evaluates `bl`'s statements and condition and *then*, only if the short circuit does not cut, `orblock`'s — byte-for-byte the CFG's own order and conditionality. `getuid()`/`geteuid()`/`getegid()`/`getgid()` are therefore safe without any purity analysis: none of them is ever hoisted, sunk, duplicated, or made unconditional. The only synthesized operation is `negate_condition_rec`, an order-neutral boolean sense flip.

**S2 — expression-shaped ops only.** Every op must be a marker (printer skips it), an op with an output, a void `CALL`/`CALLIND`, a `STORE`, or the single terminal `CBRANCH`. A `RETURN`, a second branch, a `BRANCHIND`, a `CALLOTHER` (userop rendering can be multi-line) or any no-return call declines the block.

**S3 — statement budget.** At most 2 scored statements and 2 calls per clause block. Scoring uses the same `Varnode::calc_explicit` approximation `bb_is_complex` uses and never reads `Varnode::isExplicit` — structuring runs before `ActionMarkExplicit`, so the flags are not yet meaningful.

**S4 — expression-size caps.** At most 4 condition leaves and 4 scored statements in the folded condition (`collapse_conditions` is an unbounded fixpoint loop). Every leaf of a compound operand must itself be admissible, because a nested `BlockCondition` emits *both* sub-blocks under `COMMA_SEPARATE`.

**S5 — no comment loss.** `emit_basic_block_ops` skips `emit_comment_group` entirely under `COMMA_SEPARATE`, so a clause block carrying a comment is declined outright rather than silently dropping e.g. a "WARNING: Subroutine does not return" annotation.

**S6 — no label loss.** Guaranteed by the pre-existing gates (`size_in() == 1`, `!is_interior_goto_target()`), asserted in the module docs.

**S7 — option off is byte-identical.** With `cond_join` false the admissible set is never computed, `condjoin_admits` is unconditionally false, and every gate is the verbatim upstream `is_complex` verdict.

**Declining is always correct** — the fold does not happen and today's goto+label output stands.

### Target metric table (`scripts.pipeline.compare.metrics`; angr column from the proposal branch's `angr-vs-kuna.txt`, angr 9.2.213)

| metric | angr | kuna `condjoin off` | kuna `condjoin on` |
|---|---|---|---|
| loc | 206 | 138 | **131** |
| **gotos** | **0** | **1** | **0** |
| **labels** | **0** | **1** | **0** |
| ifs | 20 | 16 | 14 |
| loops | 6 | 6 | 6 |
| ternaries | 1 | 1 | 1 |
| switches | 0 | 0 | 0 |

Success criterion (0 gotos / 0 labels with the option on, valid compilable-shaped C, `make test` unchanged with it off) — **met**.

### Speed

`python3 -m scripts.pipeline.timeit --repeat 5`: off 527.27 ms, on 518.37 ms, **delta -1.69%** (budget 5%). Within budget.

### Aggregate sanity sweep (the hard stop)

`kuna decompile-all --json` OFF vs ON over 5 binaries under `binaries/tests/x86_64/` (`cat`, `dirname`, `basenc`, `dd`, `du`), **1367 functions**; every changed body syntax-checked with `gcc -fsyntax-only` against a permissive prelude, flagging any body that parses OFF but not ON.

| | OFF | ON |
|---|---|---|
| functions decompiled | 1367 | 1367 |
| gotos | 638 | **620** |
| labels | 379 | 369 |
| loc | 34675 | 34415 |
| comma operands in conditions | 173 | 214 |

**0 invalid C, 0 crashes, 0 functions lost, 0 goto regressions.** 68 functions changed.

### decbench (bounded)

`scripts.decbench.mine --select 28` completed; of the 28 mined cases exactly **one** is touched by `condjoin`: `O2-bash-bash-get_random`, goto 1 -> 0. Rescored under the decbench venv:

```
ged_before 33.0 -> ged_after 36.0   (delta +3.0, worse)
```

Reported honestly: the fold absorbs `seedrand(); seeded_subshell = v1; v1 = brand()` into the guard as a comma operand — correct and denser, but structurally further from the source CFG, so GED penalises it (the mine also tagged this case artifact-suspect, `source_nodes = 1`). This is exactly the presentation trade-off that keeps the option **default-OFF**; no DIV is claimed.

## Gates

| gate | result |
|---|---|
| `make test` | **PARITY OK** 675/675 |
| `make test-stages` | **PARITY OK** 265/265 |
| `make rust-test` | green |
| `make check-spec` | green (lenient and `--strict`) |
| `kuna catalog --check` | `catalog OK` |

No baseline was re-pinned except `docs/baseline-stages.json`, which the new stage test requires.

## What is in the PR

* `p8_structure/kuna_condjoin.rs` — the admissibility scorer + caps + `KUNA_CONDJOIN_DEBUG=1` trace (which engine took each fold, and why a block was declined).
* The relaxed gate in `blockaction.rs` (`rule_block_or`) and `region_structurer.rs` (`try_block_or`), plus `is_condfoldable` / `cond_leaf_count` / `cond_stmt_count` in both.
* `Funcdata::has_pending_comment_at` — the in-pipeline comment lookup S5 needs (the `ArchContext` seam cannot see the console `CommentDatabase`).
* Full option wiring: `phases.toml` row, `p0_knowledge/options.rs`, both `Architecture` structs including the `build_arch_handle` copy, and all the hard-coded catalog counts (`kuna_phases/tests.rs` 76 -> 77, `catalog_bytecompat.rs` + regenerated fixture, `tests/stages/kuna-catalog.xml` #6 30 -> 31 and #7 19 -> 20, `kuna-base/src/xml.rs` corpus count 164 -> 165). Regenerated `docs/options.md`.
* Two-pass stage test `tests/stages/ghangr-condjoin.xml` (ASCII-only), self-contained gcc -O2 witness: OFF asserts the goto + its label, ON asserts the inverted compound guard.
* Spec prose in `docs/spec/08-structuring.md` under both §8.1 and §8.2.
* `docs/features/condjoin/{record.json,angr-vs-kuna.txt}`.

## What this PR does NOT close

* **Default-ON.** Not attempted, and the decbench signal argues against it: the one mined case it touches loses GED even while losing its goto. A flip would need its own ablation + DIV row.
* **The remaining loc/if gap to angr** on this function (131 vs 206 loc, 14 vs 20 ifs) is unrelated to the goto and untouched here.
* **Statement-scoring exactness.** The pre-`ActionMarkExplicit` approximation is Ghidra's own and is inexact in both directions — bash `get_random` scores 2 and renders 3 statements in its comma operand. That is readability slack, never a correctness one (`COMMA_SEPARATE` emits every op regardless); the module and spec now say so rather than claiming the score over-counts. Making it exact would mean moving the scoring after `ActionMarkExplicit`, which structuring's position in `universal_sched` does not currently allow.
* **The `branchflip` interaction** is cosmetic and unresolved: with the fold taken, `ActionBranchFlip` no longer fires on the target guard, so its `/* WARNING: branchflip: ... */` note disappears from the output. Not a comment *loss* through `COMMA_SEPARATE` (S5 covers that) — the pass simply has nothing to flip.

Closes #56.


mahaloz:
Approved!

mahaloz:
## Merge-order note — #193 (`condfold`) and #194 (`condjoin`) collide on the option count

These two PRs were developed in parallel in isolated worktrees, so neither saw the other. They are **independent** — different modules, different options, no shared logic — and **either merge order works**. But they touch the same hard-coded counters, so:

> **Whichever merges SECOND must rebase on `main` and bump every settable-option count assertion `77 → 78`, then re-record `docs/baseline-stages.json`.**

Both PRs currently move the count `76 → 77`. Merging the first is a no-op for the second's diff; merging the second without the bump fails `make rust-test` and `make test-stages` loudly (this is self-announcing, not silent), so nothing can slip through — this note just saves the rediscovery.

### Exact files the second merge must touch

| File | What to change |
|---|---|
| `decompiler/crates/kuna-decomp/src/p0_knowledge/kuna_phases/tests.rs` | `fn settable_count_is_77` → `_78`; `kuna_num_settables()` and `SETTABLE_TABLE.len()` assertions `77` → `78`; `fn tier_counts_are_16_core_37_transform_24_analysis` → `..._38_transform...` and its `(16, 37, 24)` → `(16, 38, 24)` (both options are `tier = "transform"`); `json.matches("},\n").count()` `76` → `77` |
| `decompiler/crates/kuna-decomp/tests/catalog_bytecompat.rs` | `fn fixture_has_all_77_settables` → `_78`; the `"option": ` / `"tier": ` / `"symptoms": ` count assertions `77` → `78` |
| `decompiler/crates/kuna-decomp/tests/fixtures/phase_catalog.json` | Regenerate — it is a byte-exact fixture, not a count. Procedure is in the `catalog_bytecompat.rs` module doc: `printf 'openfile write /tmp/cap.json\nphase catalog\nclosefile\nquit\n' \| decompiler/target/release/decomp_dbg -s "$PWD/specs"` then copy over the fixture |
| `tests/stages/kuna-catalog.xml` | KUNA-CATALOG **#6** (`"source_decompiler": "angr"`) `min/max="31"` → `32` and add the other option's name to the test name list; KUNA-CATALOG **#7** (`"change_kind": "structure-recovery"`) `min/max="20"` → `21` and likewise |
| `decompiler/crates/kuna-base/src/xml.rs` | The datatest corpus **file** count `assert_eq!(count, 165, "corpus file count drifted")` → `166` (both PRs add one `tests/stages/*.xml`), and extend the explanatory comment |
| `docs/options.md` | Generated — `decompiler/target/release/kuna catalog --markdown > docs/options.md` |
| `docs/baseline-stages.json` | Re-record: `kuna test --datatests --datatests-dir tests/stages --save-baseline docs/baseline-stages.json` (sanctioned — a stage test was added) |

**#194-only extra:** `kuna_phases/tests.rs` also renames `option_values_live_value_present_for_28_suppressed_for_39` → `_40`. `condfold` does not touch that test, so if #194 is the second merge it becomes `_41`; if #194 merges first, #193's rebase leaves it at `_40`.

`docs/baseline.json` (the 675-assertion datatest corpus) is **not** touched by either PR and must **not** be re-pinned by the rebase.

### ElementId — already deconflicted

Both branches originally allocated `ElementId` **4113** (`main`'s highest is 4112 = `switchreturn`). That is now resolved on the branches, so the reviewer merging either one does not need to act:

- `condfold` (#193) keeps **4113**
- `condjoin` (#194) moves to **4114**

The provenance comments on both branches now record the other's allocation, and #193's comment additionally corrects a factual error (it named 4112 as `loopbreak`; on `main` 4112 is `switchreturn` and `loopbreakrecovery` is 4092).

Nothing else in either diff overlaps.

mahaloz:
## Superseded by #196 — consolidated with #193 into one option

Closing in favour of **#196**, which carries this work forward in full. Nothing here is discarded, and the `newbury::main` criterion this PR set is met there.

### Why

This PR and #193 (`condfold`, proposal #39) turned out to be **the same feature, discovered twice**. Both relax the identical gate — the `is_complex(orblock)` decline inside `CollapseStructure::rule_block_or` and its verbatim twin `try_block_or` in the region structurer — rooted in Ghidra's `BlockBasic::isComplex` >2-statement readability budget. They collided textually too: both added a `CollapseStructure` field named `condfold_blocks`, both added a builder named `with_condfold_blocks`, this branch's entry point was even named `compute_condfold_blocks`, and both bumped the option count 76 → 77.

Two catalog entries relaxing one gate cuts against the option surface being the product, so they ship as one: `condfold off|on|wide`, ElementId 4113. The `condjoin` name and ElementId 4114 are released.

### What survived into #196

Everything, and it is load-bearing:

- `shape_score` (**Rule B**) is this PR's `score_condfold_block`, byte-identical. An independent reviewer checked every decline is present: `size_out()!=2`, non-CBRANCH tail, interior flow-break (RETURN / second branch), `CALLOTHER`, no-return call, `MAX_SHAPE_CALLS`, void op that is neither call nor STORE, `MAX_SHAPE_STMTS`, comment.
- `is_shape_foldable` / `cond_leaf_count` / `cond_stmt_count` / `shape_admits` are this PR's `is_condfoldable` / … / `condjoin_admits` verbatim, in **both** engines.
- The stage witness was **renamed, not duplicated** — `ghangr-condfold-newbury.xml`'s bytechunk is md5-identical to `ghangr-condjoin.xml`'s, all 4 assertions preserved.
- **`newbury::main` reaches 0 gotos / 0 labels at `on`**, matching angr, byte-identical before and after the consolidation.
- Your soundness argument is preserved verbatim and independently re-confirmed: `negate_condition_rec` on a `Condition` is de Morgan and **never swaps operands**, so proposal #56's worry that `getuid`/`geteuid` predicates cannot be reordered genuinely does not arise, and no purity analysis is needed.

It was also measured — not assumed — that Rule B is *necessary*: `newbury::main` folds under five `shape` folds and **zero** `prefix` folds (two of them on a nested `BlockCondition`, which Rule A can never accept). #193's predicate alone could not have closed proposal #56 at any cap.

### What got fixed on the way

- **The cap-bypass defect this PR had.** `MAX_CONDJOIN_LEAVES`/`MAX_CONDJOIN_TOTAL_STMTS` were only consulted behind `is_complex(orblock)`, and `is_complex(BlockCondition)` delegates to `getBlock(0)` alone — so once a fold produced a Condition with a non-complex leftmost leaf, a later `rule_block_or` skipped the caps entirely. #196 inherits #193's `condfolded` force-complex marking, which closes that path, plus a new unit test.
- **The cap now bounds printed width.** `MAX_SHAPE_STMTS = 2` is scored with Ghidra's `calc_explicit` approximation, which reported 2 on a block that prints **seven** statements. #196 enforces a shared printed-statement budget (5 at `on`, 9 at `wide`) in both rules; the shape allowlist is untouched.
- **The comment guard was briefly narrowed** to `comment_type::warning` during consolidation and has been restored to this PR's broader "any buffered comment" form.
- **Prose accuracy:** it is now stated outright that this is *not* a monotone goto reducer (measured regressions: `clientloop.o::process_escapes` 26→28, `true_ubuntu1804::usage` 1→2 at `wide`), and the comment-loss effect under `COMMA_SEPARATE` is documented.

The decbench result you reported honestly — `bash::get_random` `ged 33.0 → 36.0`, `source_nodes: 1`, artifact-suspect — was re-checked and stands unchanged. It remains a reason the option ships default-off.

All measurements were re-taken against `main` @ `d09f21ce`; the six-PR IDA-parity stack landed in between and changes emitted C, so every number in this PR is stale.

Branch `feat/condjoin` is left in place for reference.

PR #208, opened 2026-07-31
feat(cli): add speed-first fast mode

mahaloz:
## Summary

- add a native fast mode for decompile, decompile-all, decompile-project, functions, and the interactive console
- apply exactly listing=off, funcstart_patterns=off, and aif=off
- preserve last-write precedence so an explicit option can selectively restore analysis
- document coverage tradeoffs and add engine, CLI, catalog, and stage coverage

This is an explicit preset over existing options. It changes no defaults, catalog rows, or option counts.

## Why these three options

Listing removes the dominant whole-image decode/xref tier. Function-start patterns must also be disabled independently: on the private i386 PE, enabling funcstart_patterns while Listing and AIF remain off expands the inventory from 693 to 4,452 entries. Explicitly disabling AIF suppresses batch-driver injection and keeps it off when Listing is selectively restored.

Loader correctness passes and all shipped per-function transforms remain enabled. The tradeoff is fewer program-wide facts and, on stripped non-x86-64 inputs, potentially fewer discovered functions.

## Private-binary benchmark

Private binary SHA-256: bc4c15d826aaebeace3fec6360eb687e5662cba8745605093254931dcdb3ae1b

The old default decompile-project run had not completed after 935.91 seconds. Three release runs with --mode fast completed in 62.85 / 64.13 / 64.87 seconds, for a 64.13-second median and 141,548 KiB median peak RSS. This is an observed lower-bound speedup of 14.6x and at least a 93.1% wall-time reduction.

The C, header, and assembly artifacts were byte-identical to the equivalent three explicit options:

- C: 8aeeab9f635f4463b4aa3cf98e3db1cf135ff64023dcb8e10de1df7e96852329
- H: e556fa6ea4baa6dfef4497d1abc964c16dfcb6462fe9ebb8566f8321b191b31d
- ASM: 7b937c7fad085501141f09bd76ff371502a2cf63e6b6053a0d9142588c6aeba0

This measurement isolates the mode. Executable-only batch target filtering is handled independently in #206.

## Validation

- make test: PARITY OK, 675/675
- make test-stages: PARITY OK, 279/279
- make rust-test: green
- make check-spec: green
- independent review: no blockers

WASM does not currently expose modes or arbitrary options; adding that public API is intentionally out of scope for this PR.

### DIFF
--- a/decompiler/crates/kuna-base/src/xml.rs
+++ b/decompiler/crates/kuna-base/src/xml.rs
@@ -1659,8 +1659,8 @@ mod tests {
         // -> `?:` ternary, decbench F5 O0-iproute2-ip-print_link_flags,
         // and ghangr-switchreturn / switchreturn, the continuation of earlyreturn to the
         // WIDE multi-way switch-phi return, O0-libedit-tty__getcharindex,
-        // and modes-aggressive / the `mode reliable|aggressive` preset axis
-        // (Architecture::apply_mode; returndup-only warning as the discriminator),
+        // and modes-aggressive / the `mode reliable|aggressive|fast` preset axis
+        // (Architecture::apply_mode; catalog state and returndup warning as discriminators),
         // and ghangr-iteexpr / the `iteexpr` computed-arm ?: extension of iteregion
         // (angr ITERegionConverter over computed arms; ternary in the on-pass only),
         // and the THREE `condfold` witnesses for the short-circuit fold across a
--- a/decompiler/crates/kuna-cli/src/decompile_all.rs
+++ b/decompiler/crates/kuna-cli/src/decompile_all.rs
@@ -3,9 +3,11 @@
 //!
 //! ```text
 //!   kuna decompile-all <binary> [--json] [--functions a,b,..] [--addr 0xVMA].. \
-//!                       [--no-vars] [--max-fn-seconds N] [--option N V].. \
+//!                       [--no-vars] [--max-fn-seconds N] [--mode MODE] \
+//!                       [--option N V].. \
 //!                       [--slice ARCH] [--target T] [--sleighpath D]
-//!   kuna functions <binary> [--json] [--slice ARCH] [--target T] [--sleighpath D]
+//!   kuna functions <binary> [--json] [--mode MODE] [--slice ARCH] [--target T]
+//!                  [--sleighpath D]
 //! ```
 //!
 //! Unlike `kuna decompile` (which spawns a fresh `decomp_dbg` subprocess **per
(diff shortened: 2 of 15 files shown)


mahaloz:
All-six-PR end-to-end composition is complete on the private binary (SHA-256 `bc4c15d826aaebeace3fec6360eb687e5662cba8745605093254931dcdb3ae1b`).

The clean benchmark tree cherry-picked #201, #203, #206, this PR, #210, and #212 with no conflicts.

| Mode | Wall time | Peak RSS | Result |
|---|---:|---:|---|
| default control | >618.60s | 717,592 KiB | terminated incomplete; no artifacts |
| `fast` run 1 | 4.25s | 118,060 KiB | 351 ok, 0 failed |
| `fast` run 2 | 3.64s | 118,704 KiB | 351 ok, 0 failed |
| `fast` run 3 | 3.66s | 118,712 KiB | 351 ok, 0 failed |
| `fast` run 4 | 3.79s | 118,068 KiB | 351 ok, 0 failed |
| `fast` run 5 | 3.39s | 118,064 KiB | 351 ok, 0 failed |

Fast median: **3.66s**, 118,068 KiB. That is:

- more than **169x faster** than the same stacked default control (`>99.41%` less wall time)
- more than **255x faster** than the original-main default control, which was still incomplete at 935.91s (`>99.61%`)
- **17.52x faster** than this PR's standalone 64.13s fast-mode median once the complementary target-filtering and allocation PRs are present

All five runs emitted identical artifacts, and the C/H/ASM bytes exactly match the previously preserved executable-only 351-function export:

- C: `e1d29d6e31eeff3847f9124c07a6ba7c45adb0f1558768549b63bab79382a4c6`
- H: `dd8bfd255d009bcd9374ba4ddf58d065ddce416a2e5f1c5dfb287e0d8d8c6b9f`
- ASM: `23c5dcbeaaf32610cf055dd9f13deb92015912a93257fd78a4c0c9e52fdf3871`

The combined result validates `fast` as the practical project-export mode for this workload while retaining default mode for maximum discovery coverage.

mahaloz:
Inventory follow-up: reproducing the two project option sets with `kuna functions` makes the default/fast difference explicit.

- default project analysis (`listing on`, `funcstart_patterns on`, `aif on`): 14,014 canonical inventory entries, 10.41s discovery, 575,424 KiB RSS
- fast's explicit option triple: 693 inventory entries, 1.76s discovery; #206 then selects the 351 executable project targets

This explains why the stacked default exporter remains CPU-bound past ten minutes even after the assembly improvements: it attempts broad speculative coverage over roughly twenty times the inventory. Fast mode intentionally chooses the loader-backed executable set instead.

PR #210, opened 2026-07-31
perf(sleigh): recycle parser-context arenas

mahaloz:
## Summary

- replace per-translation SLEIGH parser-context construction with a bounded, engine-local reuse pool
- retain parser arenas while completely resetting logical state before each decode
- keep main, `inst_next2`, and delay-slot contexts distinct while live
- return checked-out contexts on success, decode errors, and unwinding through an RAII guard
- extend the deep ARM arena regression to verify repeated assembly and p-code after shallow decodes
- document the scratch-storage contract in the P2 specification

This is a strict allocation optimization. It does not cache decoded instructions, addresses, bytes, or painted context, and it changes no option or default.

## Root cause

Private binary SHA-256: `bc4c15d826aaebeace3fec6360eb687e5662cba8745605093254931dcdb3ae1b`

The project assembly sweep resolves about one million instructions. Before this change, every resolution constructed a 64-node `ParserContext`; every node independently allocated a 20-slot operand vector. Counting-allocation profiling attributed 65.71 million calls and 35.20 GB of cumulative allocation to those two objects alone.

The pool keeps up to eight idle contexts. Each checkout resets addresses, parse state, context words, commit records, the root, and every child as it is allocated. Removing a context from the pool for the full lifetime of a resolution preserves nested-decode behavior.

## Private-binary benchmark

The isolated benchmark calls the same `build_asm` path as `decompile-project`, with no function labels. Current main (`cd0f39ac`) and this commit used release builds, the same host/specs/input, five runs each, and byte-identical 60,411,921-byte / 1,010,978-line outputs.

- internal assembly sweep: median 6.698s -> 2.096s (**68.71% lower, 3.20x faster**)
- external bootstrap + sweep + write: median 8.17s -> 3.82s (**53.24% lower, 2.14x faster**)
- allocation/reallocation calls: 90,033,227 -> 22,298,036 (**75.23% fewer**)
- cumulative allocated bytes: 35,723,449,474 -> 509,238,794 (**98.57% fewer, 70.15x lower**)
- parser-arena allocations: 1,010,974 -> 1
- operand-vector allocations: 64,702,336 -> 64

All five assembly outputs had SHA-256 `2070af123ab971168895002e46409db33a0395d3d6c7f04f53d942afcdfa14d6`.

The isolated-options full command (`listing=off`, `funcstart_patterns=off`, `aif=off`) moved from a 61.45s median to 60.10s across three runs on the noisy shared host. Every run retained 693 targets / 594 successes / 99 failures, and every C, header, ASM, and README artifact was byte-identical. Decompiling all 693 targets dominates that control; executable-only automatic batch targeting remains the independent change in #206.

## Validation

- `make test`: PARITY OK, 675/675
- `make test-stages`: PARITY OK, 278/278
- `make rust-test`: green
- `make check-spec`: green
- exact private C, header, ASM, and README hashes unchanged
- independent review: no blockers

Address-result caching and parallel decoding are intentionally out of scope. The former changes invalidation requirements; the latter needs one stateful SLEIGH engine per worker rather than shared parser scratch.

### DIFF
--- a/decompiler/crates/kuna-sleigh/src/sleigh.rs
+++ b/decompiler/crates/kuna-sleigh/src/sleigh.rs
@@ -11,8 +11,9 @@
 //!   (ADR 0001): `ConstructState` nodes live in `ParserContext::state`, child
 //!   links are `Option<usize>` indices, the `ParserWalker` carries a node
 //!   index plus the breadcrumb path.
-//! - [`PcodeCacher`] / [`DisassemblyCache`] / [`SleighBuilder`] / [`Sleigh`]
-//!   from `sleigh.{hh,cc}`.
+//! - [`PcodeCacher`] / [`SleighBuilder`] / [`Sleigh`] from `sleigh.{hh,cc}`.
+//!   Parser contexts are recycled between decodes while retaining their state
+//!   arenas, the allocation-saving part of the C++ `DisassemblyCache`.
 //!
 //! The walker implements the `SymbolWalker`/`SymbolWalkerChange`/
 //! `PatternExpressionContext` hooks (slghsymbol.rs / slghpatexpress.rs):
@@ -28,6 +29,7 @@
 //! re-enters a cache mid-borrow).
 
 use std::cell::RefCell;
+use std::ops::{Deref, DerefMut};
 use std::rc::Rc;
 
 use kuna_base::address::{calc_mask, Address};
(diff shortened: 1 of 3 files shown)


mahaloz:
Stacked composition with #212 has now been measured on the same private binary (SHA-256 `bc4c15d826aaebeace3fec6360eb687e5662cba8745605093254931dcdb3ae1b`).

Over the full 1,010,974-instruction assembly sweep, five release runs produced a 1.322880s internal median: 5.06x faster than main and 1.58x faster than this PR alone. Allocation/reallocation calls fell from 90.03M on main to 6.48M, while cumulative allocated bytes fell from 35.72 GB to 197.38 MB. All outputs were byte-identical to main (`2070af123ab971168895002e46409db33a0395d3d6c7f04f53d942afcdfa14d6`).

The result confirms the changes are complementary: this PR removes parse-arena churn, while #212 removes the transient strings/vectors that become visible afterward.

PR #227, opened 2026-08-01
[AUTOMATED] fix(p3): make the opcode -> TypeOp seam total (24 functions decompiled to nothing)

mahaloz:
## The bug: a function that decompiles to nothing

24 functions in a 20,246-function desktop-Linux sweep produced **no output at
all** -- the caller got one error record and zero C. Among them: coreutils
`timeout`/`split` `main`, bash `wait_for` (+6 more), 3 tar functions, and one
each in grep, scp, sftp, sftp-server. Every one of those is a guaranteed zero in
any benchmark.

P3 carried **six hand-maintained copies** of the `glb->inst[opc]` opcode ->
TypeOp mapping, one per rule-batch file, each a partial whitelist of "the
op-codes this batch emits" with a `panic!` (or a silently-wrong flagless)
default arm. The copies drifted apart, so a rule that legitimately retyped an op
to an op-code its own file had not enumerated unwound the whole decompilation
(caught per-function, reported as an error record).

| Seam | Missing op-code(s) | Symptom |
|---|---|---|
| `ruleaction_2::op_typeop` (`panic!`) | `CPUI_INT_SRIGHT` | `RuleBitUndistribute` on `(V s>> c) & (W s>> c)` -- 24 functions / 27 panic events |
| `ruleaction_3::opflags_for` (`panic!`) | `CPUI_FLOAT_INT2FLOAT` (x86-64), `CPUI_FLOAT_LESS`, `CPUI_FLOAT_ADD` (ARM) | `RuleMultiCollapse` collapsing a phi to a float op |
| `ruleaction_8::resolve_typeop` (`panic!`) | everything outside its 20 listed op-codes | latent |
| `ruleaction_5`/`_6`/`_7` (silent fallback) | everything outside their lists | wrong/zero property bits on the retyped op |

The divergence was the tell: `ruleaction_3` *did* list `INT_SRIGHT`,
`ruleaction_2` listed things `_3` missed. Six partial copies of one total
mapping.

## The fix: derive every seam from the one total table

The tree already had the total table -- `p5_types/typeop.rs (type_op_info)`,
transcribed field-for-field from upstream `typeop.cc`, whose `match` has **no
wildcard arm** (so a new op-code cannot enter the enum without the compiler
demanding its row). `ruleaction_1`/`_4` already resolved through it; the other
six now do too, via the new `seam_type_op_for`:

* total over every op-code the enum defines;
* the one value with no upstream `inst[]` record -- the `CPUI_MAX` sentinel,
  which is not an operation -- degrades to a property-less skeleton instead of
  aborting the function (no `panic!` left in the seam);
* 291 lines of duplicated whitelist deleted.

## Is it a behavior change? No -- measured

For every op-code the old whitelists listed, the canonical table carries the
**same `opflags`**; only the debug display name differs, and that string is read
solely by the `infra/graph.rs` dev dump (`funcdata_printraw` already reads names
from the canonical table).

Sweep of 12 desktop/embedded ELF binaries, **15,596 functions**,
`decompile-all --json --mode reliable`, before vs after:

| Binary | error records before -> after |
|---|---|
| O0 coreutils timeout | 1 -> 0 |
| O0 coreutils split | 1 -> 0 |
| O0 coreutils ls | 0 -> 0 |
| O0 bash | 7 -> 0 |
| O0 tar | 3 -> 0 |
| O0 grep | 1 -> 0 |
| O0 openssh scp / sftp / sftp-server | 1 -> 0 each |
| O2-noinline coreutils ls | 1 -> 0 |
| O2-noinline coreutils timeout | 0 -> 0 |
| O2-noinline betaflight STM32F405 | 4 -> 2 |
| **TOTAL** | **21 -> 2** |

* 19 functions recovered (all of them whitelist panics).
* The 2 remaining are the `p2_lift/jumptable.rs:1811` panic and one loader read
  (`Unable to load 512 bytes`) -- both owned by other PRs.
* **Zero** functions that already decompiled changed their emitted text
  (per-function SHA of the emitted C compared across all 15,596).

Speed (median of 3, whole-binary `decompile-all`, budget +5%):

| Binary | before | after | delta |
|---|---|---|---|
| coreutils timeout | 1362 ms | 1406 ms | +3.23% |
| grep | 6946 ms | 7094 ms | +2.14% |
| openssh sftp-server | 5255 ms | 5110 ms | -2.78% |

The positives are the previously-abandoned functions now decompiling in full.

No option: this only corrects wrong output (a function that produced nothing now
produces C), and it restores upstream behavior rather than diverging from it, so
there is no DIV row either.

## Durability -- the part that stops this recurring

* `p5_types/typeop/tests.rs`: iterates **every** `OpCode` the enum defines and
  asserts the seam answers for all of them with real property bits equal to
  `inst[]`, plus the `CPUI_MAX` degradation and the four op-codes that were
  missing. Fails if any future op-code is unanswerable.
* `ruleaction_2`/`ruleaction_3` each carry the same sweep over their own local
  seam function, so re-introducing a per-file whitelist fails immediately.
* `tests/stages/ghdec-opcode-seam.xml` -- end-to-end `bytechunk -> lo fu ->
  decompile -> print C` on the minimized shape `(a >> 3) & (b >> 3)`:
  * pre-fix (verified by stashing the fix and rebuilding):
    `thread 'main' panicked at ruleaction_2.rs:96: ruleaction_2 op_typeop:
    unhandled opcode CPUI_INT_SRIGHT` and an empty body -- 2 of 3 asserts FAIL;
  * post-fix: `return (int4)(a1 & a0) >> 3;` -- all 3 pass.

## Gates

| Gate | Result |
|---|---|
| `make test` | **PARITY OK (675/675)** -- `docs/baseline.json` untouched |
| `make test-stages` | **PARITY OK (308/308)** -- stages baseline re-recorded for the new case |
| `make rust-test` | green |
| `make check-spec` | green |
| `kuna catalog --check` | green |

Spec updated in prose: `docs/spec/03-ssa-and-simplification.md` §3.2 now
describes the retype-an-op-mid-rule seam and why its totality is load-bearing.

Provenance: `inspiration = "decbench:cases-missing pool; coreutils timeout/split
main, bash, tar"`.


mahaloz:
### Scope note: one seam deliberately left alone

`p3_dataflow/heritage.rs (typeop_skeleton)` is a *seventh* local opcode -> TypeOp
builder with a `panic!` arm, and this PR does **not** touch it. Two reasons:

1. It is not reachable: all 19 call sites pass one of three literal op-codes
   (`CPUI_MULTIEQUAL`, `CPUI_SUBPIECE`, `CPUI_PIECE`), so no drift is possible
   and it produced none of the observed failures.
2. Migrating it would be a real behavior change rather than a bug fix: its
   MULTIEQUAL row carries only `marker`, while the canonical `inst[]` row is
   `special | marker | nocollapse`. Handing heritage's phis the upstream flag
   word affects `is_special()`/`is_nocollapse()` on every phi in every function
   and belongs in its own measured PR, not here.

Everything in this PR is flag-for-flag identical to what the seams already
installed for the op-codes they listed -- which is why the 675-assertion corpus
and 15,596 already-working functions are byte-identical.

PR #229, opened 2026-08-01
fix(p4): recover stack-passed call arguments (callsitestackargs)

mahaloz:
Increment 1 of the approved `[PROPOSAL]`. **One mis-ported argument**, plus option scaffolding, a two-pass stage test, the spec paragraph, and a history record. Rebased on `c99f63e3`; the ablation was re-run after the rebase.

## The defect

kuna dropped **every stack-passed call argument** at call sites whose callee prototype is unlocked — x86-64: everything past the 6th register argument; i386: every argument. Where a dropped argument was the sole consumer of a computation, dead-code elimination then deleted real basic blocks.

| decbench case | before | after |
|---|---|---|
| `O0-mydoom-mydoom-msg_b64enc` (i386 PE) | the entire `for(;;)` read/encode loop deleted — 3 basic blocks survive of a 22-node source CFG, `ReadFile` called with **zero** arguments | full `do { … } while` loop restored, `(*dat_411350)(v10,v6,0x400,v8,0)` with `v8 = &v4` |
| `O2-noinline-openssh-portable-ssh-add-parse_dest_constraint` (x86-64) | `sub_28730("ssh-add.c","parse_dest_constraint",0x2f3,1,6,0)` — 2 `if`s; the eight NULL-default ternaries DCE'd | full 16-argument call, 6 `if`s — structurally identical to IDA and Ghidra, which score GED 0 |

This is a correctness failure: kuna emitted code that does not implement the binary.

## Root cause — a mis-ported predicate

`check_input_trial_use` handed the trial's **callee**-relative address to `FuncProto::get_local_range().in_range()`, where upstream `fspec.cc:5618` hands the argument Varnode's **caller**-relative address:

```rust
// before
} else if !data.get_func_proto().get_local_range().in_range(&trial_addr, 1) {
// after (= inRange(vn->getAddr(),1))
} else if crate::p4_calls::kuna_callsitestackargs::outside_caller_local_range(
    data.get_arch().callsite_stack_args,
    data.get_func_proto().get_local_range(), &vn_space, vn_offset, &trial_addr,
) {
```

`guard_calls` registers the trial at `addr - stackoffset` while creating the Varnode at `addr`, so on a downward-growing stack the trial offsets are always positive (8, 16, 24, …) and the caller's `localrange` is negative. Every stack trial failed, was scored `no-use`, and — being *definitely* unused — had its CALL input replaced with constant `0`, which is what fed the argument-producing computation to DCE.

**The published triage hypothesis was disproved**, not built on. Both triage docs claimed `guard_calls` short-circuits via `tryregister = false` because `stackoffset` stays `OFFSET_UNKNOWN`. Instrumentation showed the opposite on x86-64 *and* i386: the placeholder resolves, `stackoffset` is recovered, `tryregister` is **true**, and `register_trial` + `op_insert_input` both run. The trials existed and were mis-scored one pass later.

Three things confirm mis-port rather than deliberate simplification:
1. The port comment recorded the wrong intent — *"outside the **callee's** local range"*.
2. `docs/spec/04-calls-and-prototypes.md` **already stated the correct rule** — *"outside the **caller's** local stack range"*. The code contradicted its own normative chapter.
3. Every other link is line-for-line faithful, with no `// STUB` and no `(kuna)` divergence marker: `create_placeholder`, `resolve_spacebase_relative`, the `guard_calls` `IPTR_SPACEBASE` translation, `build_input_from_trials`.

Introduced in `5cd4d518 rport/w10-callsite-args`.

## Gates

| Gate | Result |
|---|---|
| `kuna test --datatests` | **675/675 PARITY OK** — `docs/baseline.json` **unmoved** |
| `kuna test --datatests-dir tests/stages` | **316/316 PARITY OK** |
| `cargo test --release` (workspace) | 9 failures, **all verified pre-existing** by running the identical set against clean `main` (`c99f63e3`) in a throwaway worktree — release-mode `should_panic` and float golden artifacts, none touched here |
| `kuna catalog --check` | `catalog OK` |
| `make check-spec` | OK; `docs/options.md` regenerated and fresh |

## Measured

30-binary stratified decbench sweep (x86-64 / ARM Cortex-M / PE-i386, all three optimisation levels; 803 candidates enumerated), default-on vs `--option callsitestackargs off`:

| metric | off | on | delta |
|---|---|---|---|
| `calls_ge7_args` | 7 | 301 | **+294 (43×)** |
| `empty_short` | 1179 | 696 | **−483 (−41%)** |
| `short_calls` | 2421 | 1583 | −838 (−35%) |
| `total_calls` | 30017 | 30032 | +15 (+0.05%) |

i386 empty-argument-list fraction: **42.5% → 2.5%**. 60 runs, zero failures, zero timeouts.

## ⚠️ Speed — over budget on i386, flagged not absorbed

Interleaved, 9 repetitions, `taskset`-pinned:

| binary | arch | median delta |
|---|---|---|
| `coreutils/tr` O0 | x86-64 | −0.74% |
| `ssh-add` O2-noinline | x86-64 | +0.27% |
| `gzip` O2 | x86-64 | −2.28% |
| `bzip2` O0 | x86-64 | +0.94% |
| `chibios/ch.elf` O0 | ARM Cortex-M | +0.01% |
| **`mydoom.exe` O0** | **i386 PE** | **+8.18%** |
| **`minipig.exe` O0** | **i386 PE** | **+6.19%** |

Mechanically expected: on i386 *every* argument is stack-passed, so the fix recovers ~500 more arguments per binary and that is real heritage/DCE work — proportional to what is recovered, not a constant factor.

**The proposal's "no detectable delta" claim was a bad measurement** (min-of-3, non-interleaved) and is superseded. Note for whoever measures next: 5-rep runs on this host are untrustworthy — a 5-rep run put `coreutils/tr` at **+9.19%**; the 9-rep pinned run put it at **−0.74%**.

`docs/improvement-pipeline.md` §4 would demote an over-budget flip to default-OFF opt-in. That table governs *feature* flips; demoting this one means shipping a default that deletes live loops. **It ships default-ON with the breach recorded rather than absorbed** — this is the one open question for review. Flipping `default` in `phases.toml` plus the `architecture.rs` reset path is a two-line change if the budget is judged to bind.

## Two pinned-text corrections — both because the OLD text was wrong

Neither is a swallowed regression; both are shown with before/after C in `analysis.md`.

**1. `tests/stages/gh6882-sparcstructret.xml` #2** — `return v1 + a0` → `return v1[0] + a0`:

```c
/* before */                         /* after */
int4 p(int4 a0) {                    int4 p(int4 a0) {
  int4 v1; // stack - 0x8              int4 v1 [2]; // stack - 0x8
  f();                                 int4 *v2; // i0
  IllegalInstructionTrap(8);           v2 = v1;
  f();                                 f(v2);        /* buffer now passed */
  return v1 + a0;                      IllegalInstructionTrap(8);
}                                      f(v2);
                                       return v1[0] + a0;
                                     }
```

`f` is a **struct-return** function; kuna was dropping its hidden return-buffer pointer along with every other stack argument, so `p` rendered `f()` and the 8-byte buffer degenerated to a scalar. Both halves of the output got more correct; only the pinned text had to follow. The XML comment records this explicitly as revision 2.

**2. `verify_dwarf_prototypes::fmt_main_return_narrows_from_the_bogus_register_pair`** pinned the literal `return v13 ^ 1;`. That function's `__libc_start_main` tail call gains its real 7th (stack-passed) argument, which declares two more locals, and `namestyle angr` numbers locals in declaration order — so the *same* expression renders `return v17 ^ 1;`. A normalized whole-body diff against clean `main` confirmed the added 7th argument is the **only** substantive change. The assertion now matches the shape (`return v<N> ^ 1;`) instead of the index; its three sibling assertions (DWARF signature, no `undefined16`, no `char[16]`) are untouched and still pin the real content.

## Blast radius, stated precisely

The defect bites every call site whose callee prototype is **unlocked** — all of them in a stripped binary, and measurably about half even with full DWARF (`calls_ge7_args` 37 → 73 on six DWARF binaries), because DWARF locks only the prototypes it *describes*. On the short-call metrics DWARF builds move by exactly zero (68 → 68, 81 → 81). Both endpoints converge on the identical 73 with or without debug info.

## Known artifact (sequenced, not fixed here)

On i386, stale outgoing-argument slots from a previous call can be accepted as arguments of a later unprototyped call — mydoom renders `CloseHandle` with 5 arguments where it takes 1. Upstream Ghidra shares this and escapes it via PE IAT import prototypes, which kuna does not yet apply to indirect-call targets. Trading 192 empty argument lists for a handful of over-long ones is the right trade, and it is shown rather than hidden. Increments 2–4 (import prototypes, the remaining `guard_calls` stubs, DWARF stack-symbol binding) are **deliberately not in this PR**.

## Contents

- `p4_calls/kuna_callsitestackargs.rs` (+ unit tests) — the option and the predicate; the ported-core edit is one call.
- `phases.toml` `[[settable]]` row, `options.rs` registration, catalog counts (81 → 82, tier core 19 → 20), regenerated `phase_catalog.json` fixture, corpus file count 176 → 177.
- `tests/stages/ghdec-callsitestackargs.xml` — two-pass: `off` asserts the 6-argument truncation, `on` asserts both stack arguments recovered.
- `docs/spec/04-calls-and-prototypes.md`, `docs/history.md` (new "Convergences" section — no DIV row, since this converges on upstream), `docs/options.md`.
- `docs/features/callsite-stack-args/` — `analysis.md`, `proposal.md`, `record.json`, `before-after.txt` (four captured pairs incl. a DWARF control), `witness.c` / `witness2.c`, `argcount.py`.


mahaloz:
## Maintainer ruling on the speed-budget breach

Recording this explicitly because it is a deliberate deviation from a documented gate.

`docs/improvement-pipeline.md` §4 says: ablation clean + **over** the +5% speed budget ⇒ ship
**default-OFF opt-in**. This PR ships **default-ON** anyway. The reasoning:

1. **It is a correctness fix, not a feature.** The speed table governs judgment calls — a
   behavior that is a matter of taste has to earn its default. Here the current default output
   is *wrong*: calls print with arguments missing, and the code computing those arguments is
   dead-code eliminated (in `msg_b64enc` an entire `for(;;)` read/encode loop disappears).
   Shipping default-OFF would leave kuna emitting wrong C by default, which is the same
   argument that kept #228 ungated.
2. **It converges on upstream.** This is one mis-ported argument (`check_input_trial_use` passes
   a callee-relative address where upstream passes a caller-relative one). Default-OFF would
   preserve a port defect as the default.
3. **The cost is proportional to real work.** i386 passes every argument on the stack, so the
   fix recovers ~500 more arguments per binary; the extra heritage/DCE is work that should
   always have happened.
4. **The magnitude is uncertain on this host.** The author measured +8.18% (mydoom) and +6.19%
   (minipig) with 9-rep interleaved pinned runs, and separately showed a 5-rep run putting
   `coreutils/tr` at +9.19% where the 9-rep run had it at −0.74%. My own 5-rep interleaved A/B
   on mydoom gave a ~+4% median with `on` *faster* in 2 of 5 reps. The box was running several
   concurrent build agents throughout. So the breach is real in direction but its size is not
   pinned down.

The escape hatch exists — `--option callsitestackargs off` — and demoting the default is a
two-line change if you judge the budget to bind. The breach is recorded in `record.json`,
`analysis.md` and `proposal.md` §2.4 rather than being quietly absorbed.

**Credit where due:** the author's own proposal claimed "no detectable delta" and the author
then re-measured properly, found the breach, and reported it against their own earlier claim.

PR #239, opened 2026-08-01
[PROPOSAL] arm-entry-granularity: the ARM entry-recall gap is a discovery hole, not absorption

mahaloz:
**[PROPOSAL] — investigation only. No code in this diff. Do not merge until approved; approval re-dispatches implementation on this branch.**

`docs/decbench/recall-measurement.md` closed with one headline: **2,026 of the 2,061 live recall misses (98.3%) are function-entry granularity on embedded ARM** — "kuna emits a body for code *containing* the DWARF `low_pc` but does not start a function *at* it", median 24 bytes away.

The count holds. The mechanism sentence does not, and the difference decides the fix.

## The correction that matters

**It is not absorption — it is a hole.** For every missed address, is it inside the ground-truth body of the nearest lower kuna entry?

| | count |
|---|---|
| **dropped** — past the end of the nearest kuna function | **1,896** |
| unknown extent (nearest kuna entry is a kuna-invented `sub_`) | 1,040 |
| **absorbed** — strictly inside a kuna body | **3** |

Three. `nuttx` O2 `devnull_read`/`devnull_write` (8 bytes total) and `libopencm3` O2 `cdcacm` `_usbd_standard_request_device` are not merged into a neighbour — they produce **no output at all**. kuna emits `0x8003a8c`, then `0x8003aa8`; the four bytes in between are never decompiled. This is missing decompilation, not a naming or boundary problem, and it is worth exactly what the recall pool says.

## Framing: kuna is second of five here

Entry recall over 36,360 DWARF function addresses (8 ARM Cortex-M projects × 3 opt levels), re-measured live on this build against the *stripped* binaries decbench actually feeds:

| | recovered | recall |
|---|---|---|
| angr | 34,295 | 95.3% |
| **kuna** | **33,326** | **91.7%** |
| ghidra | 29,530 | 81.2% |
| binja | 29,364 | 80.8% |
| ida | 17,632 | 76.1% |

Ghidra finds only 1,859 of the 2,941 addresses kuna misses. "Name the Ghidra analyzer and port it" is the wrong frame — this is a tail against angr.

## Sub-classes, counts, and the owning pass

| sub-class | n | decbench-measured | owning pass | why the evidence was rejected |
|---|---|---|---|---|
| **pointer** | 1,671 | 1,402 | `aif::code_pointer_table_seeds` | finds the pointer word, then rejects on `is_thumb_function_prologue` + `MIN_SUBROUTINE_INSNS = 3` |
| **tailcall** | 513 | 226 | `listing/walk.rs` | the walk has no tail-call notion — every non-CALL flow target is a same-function successor |
| **called** | 297 | 146 | *not entry discovery* | 273/297 have a discovered caller; the call site sits past an unresolved Thumb `TBB`/`TBH` |
| **fallthrough** | 194 | 43 | Stages 1–2 | no reference at all, and no canonical prologue |
| **unreferenced** | 146 | 19 | — | mostly compiler-runtime aliases |
| **after-terminator** | 118 | 2 | Stages 1–2 | same |

Two shape facts decide the design: **93.3% of missed entries have no canonical Thumb frame prologue**, and **41% are ≤ 8-byte leaves** (`movs r0,#0; bx lr`, a bare `bx lr`). Every existing ARM discovery stage requires a prologue and rejects routines of ≤ 2 instructions.

Worked example for the `called` class: betaflight O0 `blackboxLogInflightAdjustmentEvent` @ `0x8031688` has **46 direct `BL`s**, all from `applyStepAdjustment` @ `0x80316b4` — whose third instruction is `tbh [pc, r3, lsl #1]`. The walk gives an indirect branch no static successor, so every case body past the switch, and every `BL` in it, is never decoded.

## Ground-truth noise: 3.4%, and the cited example is not it

`NMI_Handler` / `PendSV_Handler` / `DebugMon_Handler` were labelled "three DWARF symbols on one `b .` loop". In cleanflight they are at `0x803b060`, `0x803b062`, `0x803b064` — three **distinct** two-byte `bx lr` stubs, each with its own vector-table slot, and Ghidra, IDA, angr and Binary Ninja all recover all three. They are ordinary pointer-class misses.

Real noise: **105 alias/ICF surplus names** (`__floatundidf`/`__aeabi_ul2d`, …) on addresses already counted once — 3.4% of the name-based pool. A further 309 misses (10.5%) were recovered by no rival at all. Not worth chasing: **10–14%**.

## No existing option closes any of it

`--mode aggressive`, `addrtable`, `fast_funcdisc`, `eh_frame_full`, `operand_refs` each recover **exactly 0** additional entries, on every binary tried. `aggressive` resolves to the same `listing`+`funcstart_patterns`+`aif` set the `decompile-all` driver already injects on ARM, so it *is* the baseline. The earlier "aggressive recovers zero of betaflight's residual 277" replicates and generalises. Removing `funcstart_patterns` instead costs 6–490 entries per binary — the existing discovery is load-bearing and saturated.

## What a fix is worth (measured on a reverted scratch build)

| | ground truth | base | pointer-relax | tail-call | both + re-walk |
|---|---|---|---|---|---|
| 9 binaries | 9,802 | 8,926 (91.1%) | 9,520 (97.1%) | 9,092 (92.8%) | **9,722 (99.2%)** |

**91% of the residual gap is reachable by relaxing predicates that already run** — and the cost is precision, which is the whole of the remaining work:

| arm | new entries | ground truth | splits a real body | outside any DWARF body |
|---|---|---|---|---|
| pointer-relax | 560 | 348 (62%) | 117 (21%) | 95 (17%) |
| tail-call | 384 | 149 (39%) | **213 (55%)** | 22 (6%) |
| tail-call, stricter | 222 | 120 (54%) | 84 (38%) | 18 (8%) |

decbench scores per ground-truth function, so a false start costs nothing on GED and everything in real use. Shipping the naive relaxation would trade ~1,400 recovered functions for several hundred wrongly-split ones that the benchmark cannot see.

## Verdict: PROPOSAL, 4-PR sequence, all default-OFF first

1. **`cortexmvectors`** (SMALL, independent) — widen the Cortex-M vector-table signature. It silently no-ops on the four biggest contributors today: cleanflight/betaflight fail the `SHF_EXECINSTR` gate *and* the SRAM window (stack word `0x1000fff0`, CCM RAM; betaflight's table is linked at `0x20000000` for runtime relocation); crazyflie and nuttx fail `word[1] == e_entry` because their reset vector legitimately differs from the ELF entry.
2. **`ptrentry`** (LARGE, the prize) — pointer-referenced entries, with table-run corroboration and a terminating-routine validity check replacing the length floor. Acceptance target: ≥ 80% of new entries are ground truth, zero regression.
3. **`tailcallentry`** (MEDIUM) — needs `WalkState` to track per-function claimed instruction sets first; without containment it splits rotated loop heads.
4. **Listing-tier `TBB`/`TBH` resolution** — a separate gap, sequenced last.

Explicitly *not* proposed: a standalone PR for re-walking AIF's gap starts (`passes.rs:535` omits the re-seed Stages 2/3 do) — real, but worth +39 entries over 9 binaries and subsumed by PR 2.

## Also in this diff

Corrections to the two docs this contradicts: the mechanism paragraph in `docs/decbench/recall-measurement.md`, and `docs/missing-ghidra-analyses.md` §6's claim that `OperandReferenceAnalyzer`'s function creation is "subsumed by `s1-entry-disc`" — it is not, and that is the 1,671-entry gap. (Ghidra creates these from **instruction operands**; its data-side sibling `DataOperandReferenceAnalyzer` overrides `createFunctions` to a no-op — *"don't ever create a function from a data pointer"* — which is the precision guard kuna would need.)

Bundle: `docs/features/arm-entry-granularity/` — `analysis.md` (the map), `proposal.md` (the sequence), `rivals-vs-kuna.txt` (every raw table), `record.json`.

Gates: `make check-spec` green. No code changed, so the parity/workspace gates are structurally unaffected.

### DIFF
--- a/docs/decbench/recall-measurement.md
+++ b/docs/decbench/recall-measurement.md
@@ -80,6 +80,26 @@ tail-call get absorbed into the preceding function (`_usbd_standard_request_devi
 
 The remaining 35: 24 error records, 8 nearby-but-not-contained, 3 absent.
 
+> **Corrected 2026-08-01 by the follow-up investigation
+> ([`docs/features/arm-entry-granularity/`](../features/arm-entry-granularity/analysis.md)).**
+> The *count* holds — an independent address-based re-measurement on this build reproduces the
+> per-project distribution (libopencm3 136 and nuttx 87 exactly). The *mechanism* and the
+> *noise* claim above do not:
+>
+> - It is **not absorption**. Of 2,941 missed addresses, **1,896 sit past the end of the nearest
+>   kuna function and exactly 3 are strictly inside a kuna body**. The 24-byte median measures
+>   proximity, not containment. `devnull_read`/`devnull_write` and
+>   `_usbd_standard_request_device` are not merged into a neighbour — their bytes produce **no
+>   output at all**. This is missing decompilation, not a naming/boundary problem.
+> - The `NMI_Handler` / `PendSV_Handler` / `DebugMon_Handler` example is **not noise**: in
+>   cleanflight they are three *distinct* two-byte `bx lr` stubs at `0x803b060`/`0x803b062`/
+>   `0x803b064`, each with its own vector-table slot, and Ghidra, IDA, angr and Binary Ninja all
+>   recover all three. Real ground-truth noise is **3.4%** — 105 alias/ICF names on addresses
+>   already counted once.
+> - The dominant sub-class is not fall-through/tail-call but **code-pointer-referenced entries
+>   (57%, 1,402 of them decbench-measured)** that `aif::code_pointer_table_seeds` already finds
+>   and rejects on its frame-prologue and >2-instruction guards.
+
 ## Remaining error records
 
 240 pre-wave → **137** now, across the 191-binary sweep. The generic
--- a/docs/features/arm-entry-granularity/analysis.md
+++ b/docs/features/arm-entry-granularity/analysis.md
@@ -1,0 +1,244 @@
+# arm-entry-granularity — what the 2,026-function recall gap actually is
+
+`docs/decbench/recall-measurement.md` closed with one headline: **2,026 of the 2,061 live
+recall misses (98.3%) are function-entry granularity on embedded ARM** — "kuna emits a body
+for code *containing* the DWARF `low_pc` but does not start a function *at* it", median 24
+bytes from the containing kuna entry.
+
+That number is right. The mechanism sentence is wrong, and the difference decides the fix.
+
+All numbers below are re-measured on this build (HEAD `36641d35`) against the `stripped/`
+binaries decbench actually feeds the decompilers, with ground truth taken from the
+`compiled/` binary's DWARF. The full tables are in `rivals-vs-kuna.txt`.
+
+---
+
+## 1. First, the framing correction: kuna is winning this metric
+
+Entry recall over the whole ARM Cortex-M corpus (36,360 DWARF function addresses,
+8 projects x 3 optimisation levels):
+
+| | recovered | recall |
+|---|---|---|
+| angr | 34,295 | 95.3% |
+| **kuna** | **33,326** | **91.7%** |
+| ghidra | 29,530 | 81.2% |
+| binja | 29,364 | 80.8% |
+| ida | 17,632 | 76.1% |
+
+kuna is second of five and roughly 3,800 entries ahead of Ghidra. So "Ghidra and IDA recover
+these, name the analyzer and port it" is not the right frame: Ghidra finds only 1,859 of the
+2,941 addresses kuna misses, and misses several thousand kuna finds. The remaining gap is a
+*tail against angr*, not a deficit against Ghidra.
+
+## 2. The mechanism claim, tested: it is not absorption, it is a hole
+
+For every missed address, is it inside the ground-truth body of the nearest lower kuna entry?
+
+| | count |
+|---|---|
+| **dropped** — past the end of the nearest kuna function | **1,896** |
+| unknown extent — nearest kuna entry is a kuna-invented `sub_` | 1,040 |
+| **absorbed** — strictly inside a kuna body | **3** |
+
+Three. The 24-byte median distance is real, but it measures "there happens to be another kuna
+function nearby", not containment. Two witnesses, both from the recall doc's own example list:
+
+`nuttx` O2 `devnull_read` @ `0x8003aa0` (4 bytes: `movs r0,#0; bx lr`). The nearest kuna entry
+is `0x8003a8c`, whose real body ends at `0x8003a9e`:
+
+```c
+=== 0x8003a8c ===                       // the whole of kuna's output for this region
+void sub_8003a8c(void) {
+  sub_8004b24(); sub_8003acc();
+  sub_800db20(dat_8003b30,dat_8003b2c,0x1b6,0); // tail-call
+  return;
+}
+=== 0x8003aa8 ===                       // devnull_poll — the NEXT kuna entry
+```
+
+`devnull_read` and `devnull_write` are not absorbed into anything. Their eight bytes are
+simply **not decompiled**. Same for `libopencm3` O2 `cdcacm`: kuna emits `0x80016f4` and then
+`0x8001740`, and the whole of `0x80016b4`-`0x80016f4` — including
+`_usbd_standard_request_device` — produces no output at all.
+
+**Consequence for the campaign:** this is not a boundary/split problem that "only" costs a
+name. It is missing output, and it is worth exactly what the recall pool says it is.
+
+## 3. The sub-classes, counted
+
+2,941 missed addresses (the address-level pool; decbench counts names, see §5):
+
+| sub-class | n | share | measured by decbench | what reaches the entry |
+|---|---|---|---|---|
+| **pointer** | 1,671 | 56.8% | 1,402 | only a code-pointer word (vtable / fops struct / ISR table / literal pool) |
+| **tailcall** | 513 | 17.4% | 226 | only an unconditional `B` |
+| **called** | 297 | 10.1% | 146 | a direct `BL` — and kuna still has no entry |
+| **fallthrough** | 194 | 6.6% | 43 | only fall-through from the previous instruction |
+| **unreferenced** | 146 | 5.0% | 19 | nothing in the image |
+| **after-terminator** | 118 | 4.0% | 2 | nothing; previous instruction ends the flow |
+| condbranch | 2 | — | — | a conditional branch from outside |
+
+Two shape facts that decide the design:
+
+- **93.3% of the missed entries have no canonical Thumb frame prologue** (2,744 of 2,941).
+  Every existing kuna ARM discovery stage requires one.
+- **41% are tiny leaves** — ground-truth size <= 8 bytes (1,205; 852 of them pointer-referenced).
+  `movs r0,#0; bx lr`, `mov r0,r2; bx lr`, a bare `bx lr`. Every existing stage rejects a
+  routine of <= 2 instructions.
+
+## 4. Owning pass per sub-class, and why the evidence was rejected
+
+Discovery on a stripped Cortex-M image is
+`decompiler/crates/kuna-analysis/src/passes.rs::run_listing_consumers`, four stages feeding
+one recursive-descent walk (`src/listing/walk.rs`):
+
+| stage | code | what it seeds |
+|---|---|---|
+| 1 | `entry/mod.rs::full_pattern_starts` | Ghidra `<patternpairs>` epilogue-paired prologues |
+| 2 | `aif/mod.rs::raw_thumb_prologue_seeds` | raw `PUSH {..,lr}` / `PUSH.W {..,lr}` |
+| 3 | `aif/mod.rs::code_pointer_table_seeds` | code pointers whose target has a frame prologue |
+| 4 | `aif/mod.rs::run_aif` | fingerprint-matched gap starts |
+
+**pointer (56.8%) — evidence present, rejected by two predicates.** Stage 3 already scans every
+allocated section for Thumb code pointers and already finds these words. It then rejects them:
+`is_thumb_function_prologue` (the target must open with `PUSH`/`SUB SP`/`VPUSH`/`STMDB SP!`)
+and `check_valid_subroutine`'s `MIN_SUBROUTINE_INSNS = 3`. A 2-instruction leaf callback fails
+both. `devnull_read`'s pointer is in `.text`; `NMI_Handler`'s is at `.isr_vector+0x8`. Both are
+found and both are thrown away. The third guard, `listing.is_undefined(target)` ("never split a
+discovered function"), turns out to reject almost nothing here — see §2.
+
+**tailcall (17.4%) — the walk has no notion of a tail call.** `walk.rs` treats a CALL target as
+a new function and every other flow target as a same-function successor (lines 155-166). An
+unconditional `B` to a different function is therefore intra-function flow. In practice the
+branch source is usually itself inside the dark region, so nothing is emitted at all rather
+than being merged.
+
+**called (10.1%) — a second-order effect of unresolved switches.** These have a direct `BL`
+*and* a discovered caller (273 of 297). Worked example: betaflight O0
+`blackboxLogInflightAdjustmentEvent` @ `0x8031688` has 46 direct `BL`s, all from
+`applyStepAdjustment` @ `0x80316b4` — whose third instruction is `tbh [pc, r3, lsl #1]`. The
+walk records an indirect branch with no static successor (`walk.rs` module doc: "Indirect
+targets contribute NO static successor"), so every case body past the switch, and every `BL`
+in them, is never decoded. Ghidra, angr and Binary Ninja all recover this entry. The owning gap
+here is **jump-table resolution in the Listing tier**, not entry discovery.
+
+**fallthrough / after-terminator (10.6%)** — a static placed immediately after its neighbour,
+with no reference at all. Only a byte-pattern or a gap scan can find these; both are already
+tried and both reject on the prologue.
+
+**unreferenced (5.0%)** — 137 of 146 are found by some rival, 72 share an address with another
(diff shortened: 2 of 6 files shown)


mahaloz:
## Sequence progress (2026-08-03)

Steps 1 and 2 have shipped. Measured against this proposal's own acceptance bar (>= 80% of new entries are ground truth, zero regression), on the same harness: 16 Cortex-M binaries x 3 opt levels = 48 stripped images, 41,378 DWARF ground-truth addresses.

| step | slug | PR | status | result |
|---|---|---|---|---|
| 1 | `cortexmvectors` | #248 | **merged** | +96 GT on the decbench surface (**81.4%** precision, **0 lost**); with `listing off` +393 (**94.7%**, 0 lost). betaflight 1 -> 57 functions. |
| 2 | `ptrentry` | #255 | **merged** | recall 38,238 (92.4%) -> **39,851 (96.3%)**; +1,694 new, **1,613 ground truth (95.2%)**, **0 splits, 0 lost**. |
| 3 | `tailcallentry` | — | in flight | — |
| 4 | TBB/TBH resolution | — | not started | — |

Both default-OFF, both `PARITY OK 675/675`.

### Three corrections this proposal's own predictions needed

**1. The vector-table signature had a fourth defect not identified here.** All three shipped predicates were wrong *and* betaflight links `.isr_vector` into RAM above the flash it points at, so `harvest_vector_words`' "stop at the start of code" rule fired on slot 2 and the table looked one word long. Without fixing that, betaflight recovered nothing even with the signature widened.

**2. Table-run corroboration — the named precision guard for step 2 — is dominated and was not shipped.** Instrumented across 3,462 validated candidates: a switch table *is* a run, so it removes no additional split, and it costs **311 ground-truth functions** for 2.1pp of precision sitting entirely in the unverifiable "outside any DWARF body" bucket.

The actual precision model is **containment**:

| guard | n | ground truth | splits |
|---|---|---|---|
| validated candidates | 3,462 | 2,009 (58.0%) | 1,261 |
| + no referencing word shares the target's discovered function | 2,184 | 1,997 (91.4%) | 8 |
| + no referencing word overlaps a decoded instruction | 2,143 | 1,992 (**93.0%**) | **0** |

**3. The walk-reseed trap is real and scales.** Step 1 measured it at +44 entries for **54 ground-truth lost**; step 2 measured the same wiring at +13 for **734 lost and 974 real bodies split**. Both shipped the additive entry-facts wiring instead, which is what makes the results structural rather than heuristic — and why nothing is lost in either.

That also settles the optional AIF gap-start re-walk (`passes.rs:535`, ~+39 entries): closing it *is* a walk re-seed, so it sequences with step 3, not step 2.

### What step 3 has to clear

The tail-call arm remains the sequence's real precision risk — this proposal measures the naive version at **39% ground truth with 55% of new entries splitting a real body**, and even the stricter variant at 54% / 38%. Containment (`WalkState` tracking per-function claimed instruction sets) is the stated prerequisite, and step 2's result is the evidence that a containment model is what gets this class of change from ~58% to >90%. Step 3 is under the same hard gate: >= 80% ground truth with zero losses, or it reports its ceiling and does not open a PR.

mahaloz:
## Step 3 shipped (2026-08-03)

| step | slug | PR | status | result |
|---|---|---|---|---|
| 1 | `cortexmvectors` | #248 | **merged** | +96 GT (81.4% precision, 0 lost); betaflight 1 -> 57 functions |
| 2 | `ptrentry` | #255 | **merged** | recall 92.4% -> **96.3%**; +1,613 GT at **95.2%**, **0 splits, 0 lost** |
| 3 | `tailcallentry` | #259 | **merged** | +561 GT at **94.6%**, **0 splits, 0 lost**; stacked on `ptrentry`, Cortex-M recall **95.9% -> 97.2%** |
| 4 | TBB/TBH resolution | — | not started | the `called` class, 297 entries |

All three default-OFF, all `PARITY OK 675/675`.

### Step 3 re-measured this proposal's two variants at full corpus scale, and both are unshippable

Over all 96 Cortex-M images (43,809 ground-truth addresses), not the 48-image subset:

| arm | new | ground truth | **splits a real body** |
|---|---|---|---|
| proposal's naive arm | 39,128 | **1.7%** | **37,667** |
| proposal's "stricter" arm | 1,392 | 47.6% | 607 |
| **shipped** | 593 | **94.6%** | **0** |

### What actually worked, and why the sketched mechanism could not

The proposal called for an in-walk "already claimed by this function?" test backed by `WalkState` tracking per-function claimed instruction sets. That test is **order-dependent and answers "no" for every forward branch**, which is why it is not what shipped.

The shipped model rests on a small structural result: **a split cannot change what the walk decodes.** Since `F` (function entries) is a subset of `I` (decoded instructions) — an entry is walked, hence decoded — moving a tail-call target from the instruction worklist to the function worklist leaves the decode closure fixed. So the split is computed *after* the walk, with complete predecessor and region information, and emitted as an additive entry fact. **"Never removes an entry" is therefore a property of the wiring, not a heuristic** — which is why zero are lost, on every surface.

Containment is region crossing (the mirror of `ptrentry`'s same-region guard): the branch must leave the caller's entry-ordered region, all predecessors must be unconditional, the region must terminate, and the target must not open with a stack restore. Leave-one-out: drop region -> 514 splits; drop pure-predecessor -> 156; drop epilogue -> 12; drop terminating -> 0 but fewer entries.

Generalisation was checked, not assumed: **split-half holdout gives 99.2% with 0 splits on the held-out half**, per-project 90.3-100%, zero splits in every cell.

### Two disclosures

- **u-boot (ARM Linux, A32) has 2 splits** — +21 entries, 19 ground truth, 0 lost. It is the only non-zero split figure in the sequence. It was disclosed rather than gated away by restricting the pass to Cortex-M; a length floor would fix it and cost ~60 Cortex-M ground-truth entries, the same trade step 2 rejected.
- **The AIF gap-start re-walk (`passes.rs:535`) is measured and NOT folded in.** It loses 109 entries (87 ground truth) and splits 1,013 bodies, and gets *worse* with `tailcallentry` on: a split leaves the decode closure fixed, but a re-seed adds to it and makes other stages fail `is_undefined`. It needs its own investigation rather than being attached to any step here.

### Scoreboard

`--mode aggressive` is load-bearing for these measurements: betaflight is 533,396 B, so default `auto` resolves it to `reliable`, which builds no Listing at all (1 entry, 0% recall).

Remaining: step 4 only.

mahaloz:
[AUTOMATED]

## Step 5 measured and open: `poolentry`

| step | slug | PR | status | result |
|---|---|---|---|---|
| 1 | `cortexmvectors` | #248 | **merged** | +96 GT (81.4%, 0 lost); betaflight 1 -> 57 functions |
| 2 | `ptrentry` | #255 | **merged** | recall 92.4% -> **96.3%**; +1,613 GT at 95.2%, 0 splits, 0 lost |
| 3 | `tailcallentry` | #259 | **merged** | +561 GT at 94.6%, 0 splits, 0 lost; Cortex-M 95.9% -> **97.2%** |
| 4 | TBB/TBH resolution | — | not started | the `called` class, 297 entries |
| **5** | **`poolentry`** | **#278** | **open** | **+188 GT at 98.4%, 0 lost, 1 split** — and **4,220 phantom entries retired, 0 of them GT, 0 real bodies made unreachable** |

Over **98 stripped ARM images / 48,553 DWARF addresses**, baseline = steps 1-3 all ON
= 59,406 entries (measuring against bare `aggressive` re-banks merged work: betaflight
O0 is 5,798 vs 6,032).

| | new | GT | lost | splits | removed | GT removed | bodies unreachable |
|---|---|---|---|---|---|---|---|
| shipped config | 191 | **188 (98.4%)** | **0** | **1** | 4,220 | **0** | **0** |

Cortex-M recall **97.25% -> 97.43%**; **u-boot A32 58.35% -> 60.65%** — the first
thing in this sequence to move u-boot at all (steps 1-3 move it by zero, at 100%
precision and 0 splits here). Share of reported functions that are real DWARF
entries, Cortex-M: **77.1% -> 83.3%**. On 20 x86-64 ELFs and 8 i386 PEs: **0 entries
added, 0 removed**. 199/199 betaflight bodies byte-identical off vs on, sampled as
the immediate successors of removed phantoms. `make test` PARITY OK 675/675,
`make test-stages` 405/405.

### Four corrections this step's predictions needed

**1. The split count was a bug in the reference scan, not a property of the
mechanism.** Two independent measurements put the additive half at 8-19 splits, and a
third at 718 with a `$d` oracle; all three concluded zero splits was unreachable. The
cause: the ARM SLEIGH prints the resolved `[0x…]` form **only** for
`ldr rN,[pc,#imm]`. `vldr`/`ldrd` print `[pc,#imm]` and resolve the target in the
semantic body (`ARMneon.sinc:4337-4348`, `vldrRn`). Every pool containing a float or
a 64-bit constant therefore under-ran, and the pass planted an entry **on a pool
word** — `pidInitFilters`' "split" is `.word 0x42aa0000` (85.0f), `poly4d_eval`'s is
`.word 0x00000000`. Seven of the eight "splits" at the previous best config were data
phantoms, not code splits. Completing the literal set: **19 -> 1 split, 89.7% ->
98.4% GT**, and with bridging OFF.

**2. "Additive is safe, subtractive cannot be made safe" is half wrong, in both
directions.** Unpaired suppression removes 0 ground-truth *entries* and destroys
**531 real bodies** — the bar counts addresses, so it scores that as zero harm.
Pairing (drop a phantom only when its pool end carries a replacement entry) makes the
removal a **MOVE**, and bodies-destroyed goes 531 -> **0** while still retiring 4,220
phantoms. That restores a wiring-level guarantee to the half that supposedly could
not have one. Conversely the additive half is *not* free: at every bridged
configuration it destroys bodies through the suppression half it feeds (62 at 1-word
bridging alone, 119 combined). Bridging is off.

**3. This step's prize is precision, and the recall number is small.** 19 of the
charter's 324 IDA O0 misses; +17 bodies kuna could not previously emit at any
address. 171 of the 188 new ground-truth entries were already being emitted two bytes
early — an address correction, not a recovery. The number worth quoting is the other
one: **one in nine functions `kuna functions` reported on this corpus began inside a
literal pool, and not one was real.** decbench cannot see any of it, which is exactly
why four recall-metric lenses found a precision defect.

**4. The `$d` mapping-symbol oracle systematically over-states the damage.** It calls
the whole `$d` run a pool; reference-driven inference with a complete literal set
calls only *referenced* words a pool. Same corpus, same images: 73.6% GT / 718 splits
under the oracle, **98.4% / 1** under the implementation. The oracle is the right
instrument for sizing the phantom population (0 of 6,128 are ground truth, which
replicates: 0 of 4,220 here) and the wrong one for predicting a fix.

### Two disclosures

- **One split**, `SCSI_RequestSense + 4` (`0x8069c20`) on betaflight O2-noinline, one
  image. A literal reference resolves onto the function's own first word, which the
  Listing never decoded. The only guard that removes it costs 108 of 189
  ground-truth entries. Disclosed rather than gated away, the same call step 3 made
  for u-boot's two, and pinned as current behaviour by a fixture shape so a future
  fix has a test to update. The second non-GT entry is crazyflie `0x8006a30`, zero
  padding in no DWARF body.
- **Analysis-tier speed is over budget on ARM.** Only the A32 figure reproduces
  cleanly across both builds: u-boot O2 **+32.4% / +33.3%**. betaflight O0 Thumb
  measured +14.3% on the measurement build and +3.5% on the shipping build, and
  x86-64 +1.5% / +9.1% — this box carried two other agents' workspace suites
  throughout (load ~90-100), so those two rows are inside the noise floor and are
  disclosed as a range, not a point. It is net-positive only because the retired
  phantoms stop being decompiled — whole-binary `decompile-all` on betaflight O0 goes
  262.8 -> 220.7 s, a single shot. An ARM image with pools but no phantoms pays the
  cost with no benefit. Default-OFF.

### Where the rest of the gap is

**3,255 of 5,491 interior ground-truth misses (59.3%) have a containing entry that IS
a real DWARF entry** — kuna absorbed a real function into its neighbour rather than
inventing one. That is step 4's territory and it is larger than everything steps 1-3
and 5 recovered together. After this step the sequence has **not** closed ARM entry
recall; it has closed the invented-entry half of it.

One more thing for whoever takes step 4: `cortexmvectors`, `ptrentry`,
`tailcallentry` and `poolentry` all sit in `modes.rs::UNEVALUATED`, i.e. unreachable
on every default surface. They **compose** — every number above is measured on top of
the other three, and 243 of the charter's 322 cases are already exact once `ptrentry`
is reachable. Preset membership should be one joint evaluation of the four, not four
separate flips.

https://claude.ai/code/session_01C8UQbPqALzdUQ3cLLjUeKH

mahaloz:
[AUTOMATED] Closing as superseded: three of this proposal's four steps shipped, and the fourth needs re-scoping from scratch.

| step | state |
|---|---|
| 1. `cortexmvectors` | **shipped** — #248 |
| 2. `ptrentry` | **shipped** — #255 (+1,613 entries, 95.2% ground truth) |
| 3. `tailcallentry` | **shipped** — #259 (+561 entries, 94.6%) |
| — (unplanned) `poolentry` | **shipped** — #278 |
| 4. Listing-tier `TBB`/`TBH` resolution | **not started**, still a real gap |

The doc bundle here argues throughout that the fixes are unshipped, which is now false, so it is not merge-able as written; the shipped analyses live in `docs/features/{cortexmvectors,ptrentry,tailcallentry,poolentry}/`. This proposal's two genuine doc corrections — the mechanism/noise paragraph in `docs/decbench/recall-measurement.md` and the `missing-ghidra-analyses.md` §6 `OperandReferenceAnalyzer` row — were cherry-picked onto main in #345, repointed at `docs/features/ptrentry/analysis.md`.

**Scouting this proposal turned up something worth more than its remaining step**, which shipped as #345 (DIV-93): all four of those options were default-OFF and on the `UNEVALUATED` list in `modes.rs`, so they were absent from `AGGRESSIVE_OVERRIDES` and therefore **inert on every default path** (`auto` resolves to `aggressive` under 500 KiB). Measured over the 110 stripped non-x86-64 decbench twins, flipping them into the preset moves entry recall 88.63% → 93.31% (+2,373 ground-truth functions) while mid-body false entries *fall* 8,333 → 7,117, with zero ground-truth entries lost.

Step 4 (TBB/TBH) should be filed fresh against today's code if it is still wanted.

PR #285, opened 2026-08-11
[AUTOMATED] feat(p8): orchain - returndup must not split the operand chain of a short-circuit (DIV-69)

mahaloz:
## The finding

`returndup` (P8 `ActionReturnDup`, DIV-54) was flipped default-ON on an ablation over **O2
and O2-noinline only**. Extended to all three levels -- 795 slices, **85,195 scored
functions** -- it is +640 GED-perfect corpus-wide but **-192 at -O0**, and **one structural
sub-shape carries essentially all of the O0 harm and almost none of the optimized
benefit**: the split that permanently blocks `CollapseStructure::rule_block_or`.

`returndup` gives every predecessor of a shared bare-epilogue RETURN block its own private
return. For a guard clause that recovers the source shape; but the operand blocks of a
short-circuit expression are predecessors of that *same* block, and the shared out-target
is `ruleBlockOr`'s entire precondition (`bl -> {orblock, clauseblock}`,
`orblock -> {clauseblock, X}`). `ActionReturnDup` runs in fullloop's `returnsplit` group
and `collapse_conditions` runs later, so the fold is lost for good:

```c
/* option orchain off  - GED 21 */          /* shipped here (on) - GED 0 */
if (a0) return 1;                            v1 = ((a0) || ((a1 && ((a2 || (a3))))));
if (!a1) return 0;                           return v1;
if (a2) return 1;
if (a3) return 1;
return 0;
```
(`iproute2 ip::sci_complete`; the source is one boolean `return`.)

## The predicate, and its measured fidelity

The measurement identified this shape **post hoc**, by diffing the two arms' bodies for
`&&`/`||` operators the OFF body has and the ON body does not. That is an outcome, not an
implementable precondition, so **+630 was a ceiling, not a promise**, and the fidelity of a
real pre-split predicate to it is the first thing this PR has to publish.

`p8_structure/kuna_orchain.rs (shortcircuit_shared_targets)` is a read-only replay of
`collapse_conditions` on the **bblocks** CFG, run inside `returndup_apply` before the first
`node_split`. It **mirrors `ruleBlockOr`'s own admission test rather than approximating
it**: two condition nodes fuse when the sibling has exactly one in-edge, has two out-edges,
is not a switch block, is not `BlockBasic::isComplex` (the same `bb_is_complex` verdict the
rule reads, off the same op lists), and shares one out-target with the head whose other
target is not the head. Each fuse records the shared target and rewrites the head's
out-targets, so a chain collapses inside-out as the real fixpoint does. Three upstream
tests are omitted because they cannot hold at this point in the schedule (no goto has been
elected -- `select_goto` runs later; and a back edge into a single-in-edge block would make
that block unreachable). It additionally protects a block joining two or more single-exit
arms that are themselves fold targets: before the constants reach the epilogue phi the arms
are one bare `v = K` block each, which is the graph the **first** `returndup` invocation of
the fullloop sees, and without that clause the archetype recovers its `||` chain but loses
the 0/1 select `iteboolean` re-rolls.

**Fidelity to the post-hoc proxy**, over 33,804 O0 marker functions:

| | count |
|---|---:|
| proxy positives (`sc_off > sc_on`) | 2,951 |
| gate fires | 3,059 |
| **fires AND proxy** | **2,933** -- recall **99.4%**, precision **95.9%** |
| proxy AND not fired | 18 |
| **firings byte-identical to the `returndup off` body** | **2,883 / 3,059 (94.2%)** |

The 18 misses are one shape: a degenerate two-out block (both edges to the same target)
between two operands, removed later in P8 than this replay runs.

## The three-level sweep -- this implementation, not the ceiling

Base arm = shipped defaults at merge-base `82dd39a7`; variant arm = the gate. decbench's
population, names and `GEDMetric`.

| scope | n | changed | GED base | GED var | perfect base | perfect var | dPerfect | dGED |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| O0 | 32,339 | 3,014 | 230,291 | 227,475 | 14,785 | 15,396 | **+611** | **-2,816** |
| O2 | 22,546 | 1,740 | 396,062 | 397,294 | 7,422 | 7,409 | **-13** | +1,232 |
| O2-noinline | 30,310 | 2,165 | 286,500 | 287,117 | 11,129 | 11,114 | **-15** | +617 |
| **ALL** | **85,195** | **6,919** | **912,853** | **911,886** | **33,336** | **33,919** | **+583** | **-967** |

**Both directions, by name** (`docs/features/orchain/sweep-o0/moved.csv` lists every one):

| scope | to perfect | off perfect | ratio | improved | worsened |
|---|---:|---:|---:|---:|---:|
| O0 | 663 | 52 | 12.75 : 1 | 1,615 | 1,052 |
| O2 | 41 | 54 | 0.76 : 1 | 639 | 910 |
| O2-noinline | 77 | 92 | 0.84 : 1 | 1,000 | 920 |
| **ALL** | **781** | **198** | **3.94 : 1** | 3,254 | 2,882 |

McNemar z = +18.63. Per architecture **none negative**: x86-64 +530, arm +47, pe-x86 +6.
Per project 34 of 37 net-positive; the three that are not are bzip2 -3, libselinux -3,
kmod -2. Biggest moves to perfect: `iproute2 ip::sci_complete` 21 -> 0, `gzip make_ofname`
20 -> 0, `bash skipname`/`wskipname` 17 -> 0, `crazyflie sensorsFindBiasValue` 15 -> 0,
`cleanflight`/`betaflight nextArg` 14 -> 0 at **all three levels** -- that last is the
`for`-loop de-structuring DIV-54 recorded as its one net-negative cell, closed as a side
effect. Biggest move off perfect: `coreutils factor::factor` 0 -> 12, DIV-54's flagship
win, which is the honest cost: that source *did* write the guard cascade and nothing in the
binary says so.

### Against the ceiling, and against the stated falsifier

The post-hoc counterfactual recomputed on this instrument (it reproduces the measurement's
published table exactly):

| | O0 | O2 | O2-noinline | total | total GED |
|---|---:|---:|---:|---:|---:|
| ceiling -- OFF body wherever `sc_off > sc_on` | +665 | -15 | -20 | **+630** | -2,024 |
| **this implementation** | **+611** | **-13** | **-15** | **+583** | **-967** |
| (turning `returndup` off entirely) | +192 | -362 | -470 | -640 | +24,017 |

The agreed stop condition was **"below about +300 total GED-perfect, or negative at either
optimized level"**. +583 clears the first by a wide margin. On the second clause, stated
plainly rather than shaded: **both optimized levels are slightly negative, -13 and -15** --
and so is the ceiling the measurement recommended building, by more (-15 and -20). A
literal reading of that clause falsifies the recommendation itself; the decision-relevant
comparison is implementation against ceiling, where this is better at both optimized levels
and delivers 92% of the perfect count. A reviewer who wants the strict reading has the
numbers to apply it.

### Method

The two arms the corpus measurement scored are still on disk and this PR's baseline arm
reproduces its ON arm **byte for byte on 33,804 of 33,804 O0 functions**, so every function
whose gate-arm body equals one of those two takes that arm's score (exact -- GED is a
function of the body text, which is what optsweep's "unchanged bodies score identically"
control asserts); only the **312 hybrid bodies** were Joern-scored fresh. Controls: 809
unchanged-body controls reproduced their cached score with 0 mismatches, and **the O0 level
is not a projection at all** -- a full independent `scripts/decbench/optsweep.py` run over
all 265 O0 slices reproduces its row digit for digit (`docs/features/orchain/sweep-o0/report.md`),
with all three of its built-in controls passing (29,325 unchanged bodies 0 differ, 30
slug-only diffs 0 differ, baseline agrees with the published verdict on 99.8%).

### Standing-requirement-8 audit

Over all 7,018 changed bodies: **the set of called functions is identical in 7,018 of
7,018**; 2,883 of 3,059 O0 firings and 3,823 of 3,959 optimized-level firings are
byte-identical to the shipped `option returndup off` rendering. Gotos: O0 -19/+14,
optimized -12/+126 (the real cost, visible in O2's +1,232 aggregate GED).
`while( true )`: **-365 / +53** -- the gate un-destructures seven loops for every one it
creates.

## Why default-ON and not the `aggressive` preset

A new default-off option normally joins `AGGRESSIVE_OVERRIDES`. That is not enough here:
**36 of 795 slices are over the 500 KiB `auto` threshold and run `reliable`, and they hold
38,216 of the 85,195 scored functions (45%)**, including every one of the biggest winners.
Measured directly -- with the gate in the preset only, the no-option arm reproduced the
*ungated* body on `iproute2 ip` (76 functions), `bash` (366) and `openssh sshd` (201).
`returndup` is a shipped default in every mode, so its narrowing has to be one too. Hence
the DIV row.

## Speed

Whole-binary `decompile-all`, arms interleaved and pinned to the same four cores, min of 9:

| binary | off | on | delta |
|---|---:|---:|---:|
| zlib `example` (O0, x86-64) | 5,237.2 ms | 4,966.7 ms | **-5.17%** |
| coreutils `ls` (O0, x86-64) | 8,663.6 ms | 8,670.1 ms | **+0.07%** |
| crazyflie `cf2.elf` (O0, ARM Cortex-M) | 28,088.1 ms | 27,852.0 ms | **-0.84%** |
| iproute2 `ip` (O0, x86-64; min of 5, unpinned) | 48,759.9 ms | 49,252.9 ms | **+1.01%** |

Worst **+1.01%** against the 5% budget; two of four are net speedups, because the gate only
ever declines a `node_split` and the structurer then runs on a smaller graph.

## Gates (verbatim, on the rebased tree)

```
$ make test
exit: 0

=== baseline parity ===
PARITY OK

$ make test-stages
datatests: 415/415 assertions passed
exit: 0

=== baseline parity ===
PARITY OK

$ make rust-test
passed 4540 failed 0 ignored 37

$ make check-spec
check-spec OK (lenient mode)
check-spec OK (strict mode)

$ kuna catalog --check
```

675/675 is reached **with the raw flip** -- zero assertion movement, no per-test opt-out and
no re-pin of `docs/baseline.json`. The gate only ever *declines* a `returndup` split, and
the nine datatest files that pin the merged epilogue already carry `option returndup off`.

## Rebase, counts and sibling PRs

Rebased onto `11f40f46` (#280 `funcboundflow`, DIV-67). Every shared counter re-derived
from the **built artifact** on the rebased tree, never by arithmetic: live catalog 97,
`catalog_bytecompat.rs` 97, `kuna_phases/tests.rs` 97 / tiers (20, 43, 34) / `},\n` 96,
`phase_catalog.json` 97 (recaptured), `docs/options.md` 97 (regenerated), `xml.rs` corpus
195, `kuna-catalog.xml` structure-recovery 24. `docs/baseline-stages.json` re-recorded on
the merged base rather than hand-merged. **DIV number: 68 was renumbered to 69** after #277
claimed 68; re-checked against `origin/main` immediately before pushing.

Rebase drift was measured, not assumed: re-running both arms with the rebased binary over
5,777 sampled O0 functions moves **20 functions, identically in both arms** (crazyflie
`cf2`, `funcboundflow`'s own effect), so the deltas above are unaffected.

**Sequencing with the sibling P8 work.** This PR does **not** touch
`p8_structure/blockaction.rs`, so there is no textual conflict with #281 (`guardarm`) or
#283 (`loopcondhoist`); it shares `substrate/funcdata_block.rs` with #282 but in a
different function (`returndup_apply` vs `block_if_flip_negated_guard`). All four will
conflict on the shared catalog counters and `docs/baseline-stages.json`, which resolve to
base + all merged. There **is** an outcome interaction to price rather than assume: this
gate leaves a merged short-circuit condition where main leaves a guard cascade, which
changes the block shapes `rule_block_if_no_exit` and its deferred scan then see, so
whichever of #281 / #283 / this lands second should re-measure rather than add the three
deltas. Each was measured independently against shipped `main`.

The motivating measurement (`docs/decbench/returndup-o0-measurement.md`) and the
`scripts/decbench/optsweep.py` instrument both land in **#279**; this PR references them
and deliberately does not carry a copy. `docs/decbench/returndup-regression-triage.md` is
extended, not replaced -- its "no discriminator exists" and "net-positive in every
subpopulation" conclusions are now scoped to the data they were measured on, and its
Conclusion-4 loop cell is closed.

https://claude.ai/code/session_01C8UQbPqALzdUQ3cLLjUeKH


mahaloz:
## [AUTOMATED] Measurement staleness after the five-PR merge train

This PR landed last in a train of five: **#278 `poolentry`** → **#281 `guardarm`** → **#283 `loopcondhoist`** → **#284 `calloverlap`** → **#285 `orchain`**. Three of the five touch P8 block shaping, and each of their headline numbers was measured against a tree that did not contain the others. Recording here which figures survived the train and which did not — no number below was re-measured, this is a disclosure, not a correction.

### What each figure was measured on

| PR | Option | Shipped default | Headline figure | Measurement tree |
|---|---|---|---|---|
| #281 | `guardarm` | **off** | **+297 GED-perfect** over 32,339 O0 functions (post-rebase re-measure on `bb05b83e`, 139 slices / 16,019 functions) | no `loopcondhoist`, no `orchain` |
| #283 | `loopcondhoist` | **off** | **+669 GED-perfect** over 32,339 O0 functions (post-rebase re-measure on `a77e1ce5`, 139 slices / 16,019 functions) | carried `guardarm`'s *code* but with the option default-OFF, so `guardarm` was in neither arm; no `orchain` |
| #285 | `orchain` | **on** (DIV-69) | **+583 net GED-perfect / −967 aggregate GED** over 795 slices / 85,195 scored functions, all three optimisation levels (base arm = shipped defaults at merge-base `82dd39a7`; drift onto `11f40f46` measured at 20 functions moving identically in both arms) | no `guardarm`, no `loopcondhoist` |

### Which numbers are now stale, and which are not

The default-off/default-on split decides this, and it does **not** fall on "whichever landed last".

- **`orchain`'s +583 / −967 survives.** `guardarm`, `loopcondhoist` and `poolentry` all ship **default-OFF**, so none of them appears in either arm of this PR's sweep (shipped defaults vs the gate). They cannot move a number measured between two arms in which they are both inert. The one upstream default that *did* change under this PR's feet — #280 `funcboundflow`, DIV-67 — was already priced in the body above (20 functions, identical in both arms).

- **`guardarm`'s +297 and `loopcondhoist`'s +669 are the two stale figures.** `orchain` ships **default-ON**, so from `f5b83765` onward the shipped-defaults base arm those two were measured against no longer exists. This gate leaves a merged short-circuit condition where main previously left a guard cascade, and that is precisely the block shape `rule_block_if_no_exit` and its deferred scan consume — the interaction the PR body flagged in *Sequencing with the sibling P8 work*. Both arms of each of those sweeps shift, so the deltas are unverified in both directions, not merely reduced.

- **The two opt-ins are not additive with each other either.** #283's +669 was measured with `guardarm` present-but-off, i.e. as an increment over main's default rather than over `guardarm on`. `guardarm on` + `loopcondhoist on` is a third configuration nobody has measured.

### Follow-up

One combined re-measure with `scripts/decbench/optsweep.py` (on `main` since #279) against post-train `main`, over four arms — shipped defaults, `guardarm on`, `loopcondhoist on`, both on. **Do not add the three deltas**; the whole point of the interaction is that they do not compose.

None of this is a correctness question. All four gates were re-run on each rebased tree in the train and were green, and every shared counter was re-derived from the built artifact rather than by arithmetic — which caught three that git had auto-merged at the wrong value with no conflict marker. On this PR specifically: **675/675 datatest assertions PARITY OK with the raw default-ON flip and no `docs/baseline.json` re-pin**, stages 433/433 PARITY OK, `make rust-test` green, `make check-spec` OK lenient + strict, `kuna catalog --check` OK.

PR #288, opened 2026-08-12
[AUTOMATED] [PROPOSAL] p9/p5: wideintrender - valid-C rendering of 128-bit register-pair / __int128 returns

mahaloz:
See `docs/features/wideintrender/proposal.md`. From the interp-bee IDA study (finding #4): `eval_result_noun` and ~20 Bee value constructors emit `char v1[16]; return v1 << 0x40;` (invalid C). Entangled across variable-merge + type-factory (Ghidra-faithful `undefined1[16]`) + printer; three candidate gated fixes documented. **Proposal only — no implementation.**

https://claude.ai/code/session_017irbkwk5TigYwPSP7DrfPj

### DIFF
--- a/docs/features/wideintrender/proposal.md
+++ b/docs/features/wideintrender/proposal.md
@@ -1,0 +1,78 @@
+# [PROPOSAL] Valid-C rendering of 128-bit register-pair / __int128 returns
+
+Status: **proposal** (entangled across three subsystems — no implementation in this PR).
+Origin: the interp-bee/jit-bee IDA-vs-kuna study (`docs/features/funcboundflow/analysis.md`, finding #4).
+
+## Symptom
+
+`eval_result_noun @ 0xa3e0` is `mov %rdi,%rdx; xor %eax,%eax; ret` — a 16-byte
+struct/`__int128` return where `RAX=0`, `RDX=a0`, i.e. the value `a0 << 64`:
+
+```c
+undefined16 eval_result_noun(uint8 a0) {
+  char v1 [16];
+  v1[8] = 0;
+  v1[0] = a0;
+  return v1 << 0x40;   // INVALID C: array shifted; v1[0]=a0 stores 8 bytes into a char
+}
+```
+
+`array << int` does not compile, and `v1[0] = a0` silently truncates an 8-byte store
+to one `char`. ~20 `undefined16`-returning functions in interp-bee (the Bee value
+constructors, which return a 16-byte `{tag,payload}` in `RAX:RDX`) render this way.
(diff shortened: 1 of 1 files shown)


mahaloz:
Closing in favour of **#348**, which is this proposal's **Option A**, implemented,
measured and gated.

The GitHub API does not allow changing a PR's head branch (`PATCH /pulls/288` accepts
`base`, `title`, `body`, `state` — not `head`), so this could not be retargeted in place.
The work lives on `feat/arraysubfield`; `docs/features/wideintrender/` was renamed to
`docs/features/arraysubfield/` and its `proposal.md` rewritten there as the record of what
shipped, what did not, and why.

**What shipped.** Option A, unflagged. A partial access into a plain array-typed symbol
never entered `PrintC::pushPartialSymbol`, so it got a subscript chosen without reading the
access size: `v1[0] = a0` for an eight-byte store — a statement that names one `char`.
The routing gate now admits `TYPE_ARRAY`, and the walk's own `TypeArray::getSubEntry` guard
decides: `arr[3]` stays where the subscript is the whole truth, `v1._0_8_` where it is not.
Measured over 14,080 functions in 12 binaries: 308 functions / 4,953 lines change, zero
genuine in-element renders move.

**What did not, and why.**

- **Option B** (a scalar 16-byte integer type) is blocked on `p9_emit/kuna_ctypes.rs:109
  (integer_spelling)` having no 16-byte entry — it knows only `int_size` / `long_size` /
  `long_long_size`, so a scalar 16-byte `TYPE_INT` would print `uint16`, a kuna type name
  and still not compilable C. The payoff is also narrow: 147 of the 201 invalid-C functions
  are 16/32-byte SIMD blobs that are `TYPE_UNKNOWN` and must stay arrays.
- **Option C** is not new mid-end work. Its target subset is a coverage gap in the
  **already-ported** `RuleDivOpt` (`p3_dataflow/ruleaction_6.rs:2094`), which demonstrably
  fires — `sort` `sub_d040 @ 0xd040` recovers `a0 / 3` and `a0 % v3` in the same statement
  where it leaves `SUB168(v1 * ZEXT816(0xaaaaaaaaaaaaaaab),8) & 0xfffffffffffffffe` raw.
  Refiled as #343.

**Correction to this proposal.** Its headline promise — valid C — is not met and cannot be
met by Option A. `return v1 << 0x40;` survives; the scout's invalid-C detector counts
exactly the same 201 functions before and after. Also unmentioned here: `--option returnpair
single` already produces the valid-but-lossy IDA rendering for the return subcase today.

https://claude.ai/code/session_011KpG7qK6BDFZyPnyo4r1c1

PR #293, opened 2026-08-12
[AUTOMATED] feat(p8): almostregion - identify inlined code by deleting the edge the structurer virtualized

mahaloz:
**Stage 1 of two.** This PR **identifies** code the compiler inlined. The follow-up (`outline`) **removes** a chosen region and emits a call to a synthesized pseudofunction. They are split because the second is destructive and the first is not, and because the region selection can come from a human or an LLM rather than from this pass.

## The idea

SAILR (Basque et al., USENIX Security 2024) observes that a residual `goto` marks where the *compiler* restructured control flow past what a source-level `if`/`while` nest can express. kuna already exploits this three times — `gotoreduce`, `taildup`, `crossjumprevert` — each undoing a specific optimization. Inlining is another such restructuring, so the same signal should say **which bytes of this function did not come from this function**.

## Why the edge is deleted, not un-virtualized

The obvious reading — "un-virtualize the edge and re-run region identification" — is a **no-op**, and this is the first thing worth knowing about the area.

Virtualization is recorded on `sblocks` and never on `bblocks`: every `set_goto_branch` call site is on the structurer's own graph, and `KunaRegionIdentifier::build_from_block_graph` rebuilds from `bblocks` with no edge-flag filter. P7 therefore *already* sees the fully un-virtualized CFG, so un-virtualizing and re-running returns exactly the regions P7 already declined to find.

The operation that carries the intended meaning is the inverse. For each virtualized edge `e = (s,t)`: build a scratch region graph over `bblocks`, **delete** `e`, and keep a `(head, exit)` pair only when it is a single-entry region in `G − e` and **not** one in `G`. Both directions are asserted, so a set that was already a region is never reported, and the deleted edge must be **incident** to the region it exposed — without that, deletion perturbs dominance at a distance and the pass reports regions unrelated to the goto (the first implementation did exactly this).

The single-entry predicate is transcribed from the private `check_region` in `p7_regions/kuna_regionid.rs` rather than exposing it, because that file is a line-faithful port pinned by `docs/spec/07-regions.md`.

## Scheduling is load-bearing

The pass runs immediately after `ActionFinalStructure` and **before** `gotoreduce` / `taildup` / `ifelseflatten` / `crossjumprevert` / `returndup`. Those passes exist to *delete* gotos by duplication and consume **~19% of the raw virtualized-edge set**.

The stage witness is the end-to-end proof: it reports a region in a function whose final C contains **no `goto` at all**, because `returndup` duplicated the shared return tail away before printing.

## The premise, measured — including where it fails

decbench ships an oracle for this that had not been used: every project is built at both `O2` and `O2-noinline`, and the `compiled/` copies carry full DWARF including `DW_TAG_inlined_subroutine`. Measured over **1,483 paired functions** (`coreutils/tr`, `coreutils/ls`, `bzip2`, `gzip`, `tar`), with every goto-removal pass disabled:

| quantity | value |
|---|---|
| raw gotos, O2 vs O2-noinline | 3,513 vs 2,151 (1.63×) |
| gotos consumed by kuna's own cleanup passes | 650 of 3,513 (19%) |
| marginal gotos per 100 *inlined* bytes | 0.826 |
| baseline density in the `-fno-inline` build | 0.514 |
| **inlined bytes vs ordinary bytes** | **1.61× as goto-dense** |
| goto *density*, with vs without inlining | 0.671 vs 0.699 — **no separation** |

So **"there is a goto here" is not an inlining detector** — the function-level association is largely a size effect. What this pass reports is the specific structural fact underneath the goto, not the goto.

And it finds only the **messy half** of inlining. `openssh-portable/ssh::sshkey_drop_cert` at O2 has three DWARF-confirmed inlined callees and **zero** gotos even with every goto-removal pass disabled. `docs/spec/08-structuring.md` §8.5 names the signals this one cannot see — interface narrowness (measured at 2.0× separation on the same oracle), stack-slot lifetime clustering, windowed FID matching, repeated isomorphic regions.

## Regions are block sets, not address intervals

An inlined callee is routinely laid out in several disjoint pieces — DWARF records 3–6 ranges for the callees inlined into `tr::parse_str`, and in the stage witness the **exit address is below the head**. The report emits the full member list; consumers must never take `head..exit`. This matters for the `outline` follow-up: an interval-based outliner cannot work.

## Validation

On `tr::parse_str` (18 DWARF inline sites, 16 gotos at O2, 0 at `-fno-inline`) the pass reports 3 regions, all landing inside DWARF-confirmed inlined code — one a clean innermost single-callee hit, one spanning two adjacent callees, which is a real imprecision and is recorded as such. `docs/features/almostregion/dwarf-oracle.txt` shows the full pane, including the caveat that the enclosing `build_spec_list` covers most of the function so single-function overlap is weak evidence on its own.

## Option

`almostregion off|on|report`, default `off`, ELEM 4123. **Report-only** — no p-code and no structured node is touched, so `off` is byte-identical and `on` differs by comment lines alone. Standing requirement 7 (whole-corpus sweep) reduces to zero changed functions; both parity gates confirm it.

## Gates

| gate | result |
|---|---|
| `make test` | **PARITY OK** 675/675 |
| `make test-stages` | **PARITY OK** 443/443 (7 new) |
| `make rust-test` | 4,560 passed, 0 failed |
| `make check-spec` | OK, lenient **and** `--strict` |
| `kuna catalog --check` | catalog OK |
| speed | **+2.81% min / +1.59% median**, budget 5% |

Speed note: `scripts.pipeline.timeit` runs all OFF samples then all ON samples, and on this box reported **−31.94%** for a pass that cannot make anything faster — the first arm pays every cold cost. The number above is interleaved min-of-N, 13 samples per arm with alternating arm order, on the ON arm of a goto-heavy function (8 probe edges). Shipped default is `off`, where the pass returns before building any graph.

## Two count-tests worth flagging

Adding a pass to the schedule trips two hard-coded counters that are **not** in the `CLAUDE.md` checklist, and neither has an obvious name:

- `tests/fixtures/list_action_decompile_oracle.txt` — the action-tree dump is byte-compared, so a new pass shifts every subsequent line number.
- `tests/verify_w8_fw_universalaction.rs` — a hard-coded leaf count (`271 → 272`).

## Conflict with #292

This branch takes `catalog_bytecompat` from **102 → 104**: the pre-existing 102/103 drift (which #292 fixes standalone, and which had `make rust-test` red on main) plus the row added here. The two will conflict on that file; resolve to **104**. Happy to rebase after #292 lands instead — say the word.

https://claude.ai/code/session_01XTrU5w3qoWMyu38FprpfXZ


mahaloz:
## Measured against the DWARF oracle — the detector does not work. Converting to draft.

I opened this on a spot check of one function. I have now run the measurement I should have run first, and it does not support the claim.

**Protocol.** `almostregion report` over 11 decbench O2 binaries (`tr`, `ls`, `dir`, `bzip2`, `gzip`, `tar`, `find`, `diff`, `grep`, `sort`, `cp`); every reported region scored against DWARF `DW_TAG_inlined_subroutine` in the unstripped copy. Basic-block leaders derived with capstone.

- **purity** = fraction of a region's member block-starts inside one single innermost inlined callee (generous: best-matching callee, not a fixed one).
- **control** = the per-function **base rate**, i.e. what fraction of *all* that function's block-starts are inside *any* inlined range. This is the number that matters: on a function where one outer callee covers most of the body, "the region landed inside inlined code" is nearly free.

**Results — 578 regions across 197 functions:**

| metric | value |
|---|---|
| mean purity (single innermost callee) | 0.429 |
| mean "inside any inlined range" | 0.455 |
| **control: per-function base rate** | **0.417** |
| **enrichment** | **1.09× (any) / 1.03× (purity)** |
| precision @ purity ≥ 0.8 | 0.256 |
| **recall** (of 4,404 instances spanning ≥2 block leaders) | **0.045** |
| regions no better than their own function's base rate | 285/578 (49%) |

**Conclusion: the reported regions are statistically indistinguishable from picking blocks at random within the same function.** 1.03–1.09× is noise, and recall is 4.5%.

This is consistent with the premise measurement already in the PR body — inlined bytes are only ~1.6× as goto-dense as ordinary bytes and goto *density* does not separate inlined from non-inlined code at all. What I did not appreciate is that the region predicate does not rescue the weak input: conditioning on "becomes single-entry when this edge is deleted" adds essentially nothing.

My original spot check (`tr::parse_str`) was exactly the misleading case I flagged in `dwarf-oracle.txt` but then leaned on anyway: its base rate is high because `build_spec_list` spans most of the body.

### What is and isn't wrong

The pass computes what it says — the SESE predicate and the edge-deletion probe are correct, and the stage test pins real behaviour. What is wrong is the **claim** that this identifies inlined code. It doesn't.

### Disposition

- This PR should not merge as an inline-candidate detector.
- `kuna_check_region` and `build_scratch` are correct and are reused by #294 (`outline`), which is unaffected and independently measured. They will move to a neutral home so #294 keeps working without the detector or its framing.
- A replacement generator is being designed against this same oracle, with the falsification threshold stated up front this time.

Leaving the branch up for the machinery and the measurement harness; converting to draft so it is not merged on the original story.

mahaloz:
Superseded. The detector this PR shipped was measured against the DWARF oracle after the fact and scored 1.03-1.09x enrichment over the per-function base rate at 4.5% recall - indistinguishable from chance (see the measurement comment above).

`outline` (#294) is merged and carries the two helpers it needed (`kuna_check_region`, `build_scratch`) inline, so nothing here is load-bearing any more. Closing rather than leaving a disproven detector open.

The measurement harness and the oracle methodology survive and are the standing bar for any future detector: precision and recall against `DW_TAG_inlined_subroutine`, controlled by the per-function base rate.

PR #351, opened 2026-08-28
[AUTOMATED] [PROPOSAL] variantguard — pick a Rust enum's variant from the DWARF discriminant, and why the memory-object guard is not soundly reachable this way

mahaloz:
> ## PROPOSAL — nothing ships
>
> **Round 3 was refuted, and this is now a `[PROPOSAL]`.** The **memory-object
> guard direction is known-unsound**; the option is **default-OFF and excluded
> from every preset**, and default output is byte-identical to the option off.
> `docs/history.md` is byte-identical to `main` — no DIV row, because nothing
> ships.
>
> **The deliverable is [`docs/features/variantguard/analysis.md`](docs/features/variantguard/analysis.md)**,
> with the full witness corpus in
> [`docs/features/variantguard/witnesses/`](docs/features/variantguard/witnesses/).
> The implementation stays on the branch as the reference implementation.
>
> ### The refutation
>
> ```c
> v1 = (*dst).tag;                 // read hoisted ABOVE the event
> wipe(dst);                       // wipe() sets *dst = Err(99)
> if v1 & 1 != 0 { return 0; }     // branch left BELOW it
> return (*dst).payload.Ok.__0;    // WRONG: the object is Err(99)
> ```
>
> …and kuna renders `wipe` as `(*dst).payload.Err.__0 = 99;` in the same run.
> A `CBRANCH` is its block's last op, but **its condition Varnode need not be**:
> `compute_regions` raises the event-holding block's fact to ⊤ and then
> immediately re-applies that same block's *stale* edge constraint, resurrecting
> the killed fact in the successor. `hoist_xblock` puts the clobber in another
> block so an intra-block kill cannot see it at all.
>
> Four adversarial rounds each found a different shape of **one** defect — a fact
> about an enum object's discriminant applied where it no longer holds — and each
> was found only by a purpose-built fixture, never by the gates and never by the
> tests written for the previous round. Closing it needs reaching definitions plus
> an escape discipline over the object: a different piece of work.
>
> ### What is verified sound, and is where a retry should start
>
> The **producer half** (writes only, intra-block, positional, killed) survived
> every attack in rounds 2 and 3. The **value-object guard** held under every
> probe — an SSA value cannot be clobbered, so the resurrection above has no
> analogue there. Recommendation: **value-object guard + producer writes only,
> memory-object guard omitted until there is an escape analysis.**
>
> Everything below this box is the record of the three implementation rounds and
> is retained for the account, not as a merge request.

---

## Revision 3 — one kill discipline, and what it costs

A re-verification panel confirmed the revision-2 producer fix, then found the **same defect
class through the guard door**. It was pre-existing from revision 1 — the guard path is
textually unchanged since then — but it shipped as the default and it is the same lie.

### The unifying diagnosis, which is the coordinator's and is correct

Every defect this feature was sent back for is one thing: **a fact about an enum object's
discriminant applied at a program point where it no longer holds.** Revision 2 fixed that
for stores. Revision 3 fixes it for guards, and makes the two halves share one mechanism.

Three witnesses, all default output, all now closed:

```c
// [1] a write that builds Err, inside an Ok-guarded block -- BEFORE
(*dst).tag = 1;
(*dst).payload.Ok.__0 = v1 + 1;      // contradicts the line above it
// AFTER
(*dst).tag = 1;
(*dst).payload.Err.__0 = v1 + 1;     // and the guard-proved read above is still Ok

// [2] a read after a call that overwrote the object -- BEFORE / AFTER
set_err(dst, x);
return (*dst).payload.Ok.__0;        →   return (*dst).payload.field_0x8.__0;

// [3] no call at all -- BEFORE / AFTER
(*dst).tag = 1; (*dst).payload.Ok.__0 = x; return (*dst).payload.Ok.__0;
(*dst).tag = 1; (*dst).payload.Err.__0 = x; return (*dst).payload.field_0x8.__0;
```

[3] is the decisive one: `read_after_call_ok_guard` and `read_after_call_err_guard` are
identical but for which arm the guard selects, and the same load rendered `Ok` in one and
`Err` in the other — each naming the variant the preceding call had just destroyed. That is
the round-1 refutation criterion met verbatim, and it is now a committed control pair whose
two loads must render **identically**.

### The fix

**One kill discipline, shared** (`object_events`). A guard proves what the object was AT THE
GUARD; a store proves what it becomes. Neither survives an **event**:

* a **store over the tag bytes** — a payload-only store is not an event, because it cannot
  change which variant is live (and under a niche encoding the tag bytes *are* payload
  bytes, so the range test reports the overlap and it correctly does kill);
* a **call** — a callee handed the pointer may store anything through it;
* **any store the pass cannot attribute to the object**, because it may alias it.
  `*p = Ok(1); *q = Err(2); (*p).payload = x;` is a committed witness where `q` may be `p`.

A **value** object has no events at all — an SSA value cannot be clobbered. That is not a
concession, it is what SSA means, and it is why `use16` keeps its names while the same shape
behind a pointer loses them across a call.

Applied: the fact *leaving* a block holding any event is ⊤; within a block the kill is
**positional** (`Regions::guard_at`), so a read above a clobber keeps the guard and one below
it does not. A guard's own `CBRANCH` is its block's last op, so a re-test after a clobber
still constrains the edge — which is how `if (tag == 0)` below a clobber stays usable.

**And the precedence arm is deleted.** `(true, _) => here.trailing_zeros()` made a singleton
guard beat a producer fact the analysis had computed *correctly*. The two are now
**intersected**; a disagreement REFUSES and is never resolved by precedence.

### What it costs — measured, and reported as a loss

On the std-linked `-g` Rust witness the recovery drops from **5 functions / 17 labels to
4 / 11**. Every lost label is a read below a CALL: the `Tree` drop glue entirely, `depth`'s
`Node.field_0x8`, and one `Bin.__1` in each of `eval` and the `Expr` drop glue. Those are
refusals, not regressions — the honest price of not naming a value an opaque callee may have
replaced. What survives is still hand-checked against source: `eval`'s `Bin.__0` / `Neg.__0`
/ `Lit.__0`, `depth`'s `Leaf.field_0x8`, `use_divide`'s `Ok` / `Err`.

The coarsest rule is that **any call kills a memory object's guard**, with no attempt to prove
the callee cannot reach it. Refining that needs an escape analysis for the object's address —
a real analysis, not a predicate — and I did not build one. It is stated in the module header,
`phases.toml` and the spec rather than hidden.

The true positives are all intact: `use16` (`Ok`/`Err`), `ret_multi` (`P`/`Q`), and every
revision-2 refusal.

### Tests

The fixture gains six functions (guard-side): `guarded_write` and its mirror, `stale_read`,
the control pair `read_after_call_ok_guard`/`_err_guard`, and `alias_clobber`.

* **16 cross-crate tests** (was 11), five new. **Ablating the option fails 12 of 16** — every
  new test is non-vacuous, including the new control pair, whose anchor is co-located by
  decompiling `guarded_write` alongside it so it cannot be satisfied by silencing the pass.
* **22 stage assertions** (was 16), six new, each anchored against a still-named write.

### Two documentation defects, also fixed

1. **DIV-97 described the deleted whole-block analysis as the shipped design.** Root cause: in
   revision 2 an `AssertionError` aborted the `history.md` script *before it wrote*, so none of
   its edits saved and I did not re-check the file. DIV-97 is now rewritten from scratch to
   describe what actually ships, including the revision-3 kill discipline.
2. **~1,719 characters of variantguard narrative had landed in the DIV-68 row.** Same root
   cause: the follow-up script's `replace(..., 1)` hit the file's *first* `Gates:` match, which
   was DIV-68. **DIV-68 is restored byte-for-byte from `main`** (verified by `diff`), and the
   text now lives in DIV-97.

`docs/baseline-stages.json` re-verified as purely additive against `main` by key-set diff:
22 keys added, all `VARIANTGUARD`, **none removed**.

---

## Revision 2 — the blocking defect, and the fix

A five-lens review panel found a real defect with working counterexamples. **It was
mine, it shipped as the default rendering, and it is the same class of error #333 was
refuted for.** This section is the correction; everything below it is unchanged except
where marked.

### What was wrong

The producer half propagated a constant `tag = K` store **backwards over whole blocks**
with no ordering test and no kill. A store is evidence about what the object *becomes*,
so colouring every op the block reaches meant a payload READ that executes *before* the
store was labelled with the variant the store was about to write:

```c
v1 = (dst->payload).Err.__0;   // reads whatever the CALLER passed
dst->tag = 1;
```

The control is what makes that decisive rather than merely doubtful. `read_then_clobber`
and `read_then_clobber_ok` are byte-identical but for the clobber constant, and the bug
gave the **same `mov 0x8(%rdi),%rax` opposite variant names** — `Err` in one, `Ok` in the
other. Two identical reads of one source expression cannot both be right, and by
construction neither was: the source reads both arms. A second witness printed `Err` on a
value it then tested for `Ok` and returned unchanged, three lines apart.

**The shape is ordinary, not exotic.** Read-then-overwrite of an enum in place is what
`mem::replace`, `Option::take`, and any state machine that consumes its old state compile
to, and it is what drop glue and `?`-desugaring produce. The two fixtures this feature
first shipped with never read an object they later clobber — which is exactly why it
survived my own testing, and why the fixture below is now committed.

### The fix

The guard fact and the producer fact are evidence about **different things** and are no
longer mixed:

* a **guard** is evidence about the value an execution is *carrying*, so it colours every
  op its edge reaches, reads included. Unchanged — the forward edge analysis, the def-use
  walk and the printer seam are all as reviewed.
* a **producer store** is evidence about what the object *becomes*, so it now colours
  **writes only**: a store into the object, and the pointer arithmetic that feeds one and
  nothing else (`writes_only`). The fact is read **positionally** off the nearest constant
  tag store above or below the write **in its own block**, is **killed** by any
  intervening store over the tag bytes that cannot be read as a constant and by any call
  (which may store through the pointer itself), and is **refused** when the two sides
  disagree (`producer_writes`). The whole-block backward analysis, and the `UNKNOWN`
  apparatus that went with it, are deleted.

Both directions are sound for a write and both are needed: `payload = v; tag = 0;` is a
constructor whose bytes *become* `Ok`'s payload, and `tag = 1; payload = v;` writes the
payload of an object that is already `Err`. Neither is sound for a read, and no read is
named by a store any more — only by a guard.

### The panel's sharpest framing, now a test

The reviewer's `guard_then_clobber` had the pass *declining where the evidence existed and
firing where it did not*: a read genuinely proved by a dominating tag guard was left
unnamed, while the store beside it was named off the bogus block fact. After the fix both
are right, and for the right reasons:

```c
if (dst->tag & 1)                        // branch-flip
  v1 = 0;
else
  v1 = (dst->payload).Ok.__0;            // named by the GUARD  (tag == 0)
dst->tag = 1;
(dst->payload).Err.__0 = v1 + 1;         // named by the STORE  (a write)
```

### It costs nothing measured

The std-linked `-g` Rust witness changes **the same 5 functions with the same 17 labels**
before and after the fix — every one of them was consumer-side. The only whole-binary
difference between the two revisions is `std::path::Components::as_path` crossing the 10 s
per-function watchdog in one run (the same function DIV-87 reported; empty body in one
run, 1,556 lines in the other, zero variant labels either way). `ret_multi` — the producer
side's entire contribution — is unaffected, because a constructor writes its payload and
its tag in one block.

**Does the producer side still earn its complexity?** It is ~90 lines and it recovers
exactly one thing: a constructor naming the variant it builds (`ret_multi`'s `.P.a` /
`.Q.`, and the writes in the new fixture). On the real `-g` binary it contributes **zero**
— all five recoveries there are guards. I kept it because naming the variant a constructor
builds is a genuine recovery and the rule is now small and positional, but it is the half
a future reviewer could delete with the least loss, and that is worth recording.

### Also fixed (proof-channel audit)

1. **Contradiction removal is now sticky.** `(field 0, field 1, field 0)` at one key no
   longer resurrects the key; a contradiction writes a permanent `None` tombstone.
2. **`slot` is in the address key.** It was strictly coarser than the resolution it
   authorizes. The one slot rewrite the cast plane performs — a spliced zero-`PTRSUB` read
   back at slot `-1` — is now recorded as an explicit alias, and only for a pointer pin
   this pass actually made.
3. **`MULTIEQUAL` slots use their own in-edge** in the pin loop, matching the rule
   `use_contexts` already applied; one `ctx[op]` no longer covers every slot.
4. **`set_union_field`'s return is honoured** — no proof is recorded for a resolution that
   a previously locked association refused. `set_address_based_union_field`'s is
   deliberately *not* read the same way, and the reason is worth stating: `ResolveEdge::new_addr`
   discards the slot (`unionresolve.cc:118-129` encodes a fixed `0x2000`), so this pass's
   own two pins on one op collide there by construction and the second always reports
   false. Honouring it naively silenced `use16`'s `Ok` arm. It is best-effort, and a
   genuine conflict (a locked entry naming a different member) tombstones the address
   proof instead.
5. **`use_contexts` intersects over ALL matching phi slots**, not `get_slot`'s first.

### New committed fixture and tests

`decompiler/crates/kuna-analysis/tests/fixtures/variantguard_clobber_x86_64` (+ vendored
`.rs`), 5 functions covering read-then-clobber, the `Ok` control, a read kept live across
the clobber, the guard/clobber coexistence case, and a conditional store.

* **5 new cross-crate tests** (11 total), four of them refusals:
  `a_store_never_names_a_read_above_it`,
  `the_clobber_control_renders_its_read_identically` (asserts the two reads render
  *identically*, which is the decisive form), `a_read_kept_across_a_clobber_is_not_named`,
  `a_conditional_store_proves_nothing_about_the_read_above_it`, and
  `a_guard_proved_read_coexists_with_a_clobber_of_the_other_variant`.
* **6 new stage assertions** (16 total), including a count that ties the unguarded pair and
  the control together so a future regression cannot pass by breaking only one of them,
  and `#15`, which pins that the *writes* are still named — so the refusals cannot be
  satisfied by silencing the pass.

### Two secondary findings, dispositions

* **`use16` renders `v2.payload.Err` while `put_res` renders `(dst->payload).field_0x8.__0`**
  — different descent depths for the same kind of access. **Not introduced here**: with
  `variantguard off` those same two functions already render `v2.payload.field_0x8` and
  `(dst->payload).field_0x8.__0`. It is the implied-field seam (one member token) versus
  the `PTRSUB` seam (full field path), i.e. DIV-87's own rendering. This pass substitutes
  the member NAME at whatever depth the printer chose and never changes the descent.
* **`ret_multi` prints `.P.field_0x8` where DWARF says `P.b`** — real, and left as the
  disclosed incompleteness. Renaming the field inside a proved facet needs a third type
  index plus the same treatment at two more printer seams; doing it at one seam and not
  the other would manufacture exactly the inconsistency above. Deferred, and stated in the
  module header, `phases.toml` and the spec.

`scripts/wt_setup.sh` has been dropped from the branch.

---

## What this is

**The dominating-guard analysis #334 recorded its side table for.** `dwarfvariants` (DIV-87)
imports a Rust tagged enum's `DW_TAG_variant_part`, installs it as
`struct { tag; union payload; }`, records the exact geometry — the discriminant's offset and
width, and per variant its source name, its `DW_AT_discr_value` and its payload struct — and
then **refuses to name a variant** wherever two of them overlay one byte range. A union member
selects itself by OFFSET and the discriminant is never consulted, so `Ok.__0` and `Err.__0`
(same offset, same width) tie in `ScoreUnionFields` and the winner is trial order. Every
`Result<T,E>` is that case.

#334's module header, `kuna_dwarfvariants.rs:111` and `docs/spec/01-program-prep.md` all said
the same thing: *"Picking the facet from the tag needs a dominating-guard analysis; that is
what the side table is recorded for."* This PR writes it, and it is sound by construction: the
compiler states which value selects which variant; only the block→value mapping is derived.

```c
// dwarfvariants alone (DIV-87)                    // + variantguard (this PR)
v2 = r16(x);                                       v2 = r16(x);
v1 = (int8)v2.payload.field_0x8 + 100;             v1 = (int8)v2.payload.Err + 100;
if (!(v2.tag & 1))                                 if (!(v2.tag & 1))
  v1 = v2.payload.field_0x8;                         v1 = v2.payload.Ok;
```

Source: `match r16(x) { Ok(v) => v, Err(e) => e + 100 }`. `option variantguard on|off`, **P5**,
tier `transform`, **default ON**. Spec: `docs/spec/05-types.md` §5.4 + `docs/spec/09-emission.md`
§9.2. DIV-97.

## The two seed kinds — I shipped BOTH

Both are anchored on a Varnode whose **recovered type** is the enum (or a pointer to it), so no
codegen shape can trigger either — which is the premise #333 was withdrawn for.

* **Producer** (a constant store over the tag bytes). Fires on `ret_multi`, and agrees with the
  already-correct labels on `ret_three` / `list_len` / `ret_niche`.
* **Consumer** (a `CBRANCH` testing the tag). **This is the bigger prize and it works.** The
  predicate is evaluated over the enum's *discriminant domain*: every `DW_AT_discr_value`, plus
  the DEFAULT (niche) variant, admitted only when a value **outside** the claimed set satisfies
  that edge — counted exactly (`TagPred::population`), not assumed. That exactness is what makes
  `if (p == NULL)` on an `Option<&T>` resolve to `None` on one edge and `Some` on the other
  instead of admitting both. Masked (`test al,1`), truncated, zero-extended and negated forms are
  all read.

**The hoisting problem, and why a block region is not enough.** rustc at `-C opt-level=1`
computes `Err(e) => e + 100` *before* the branch and selects it with a `cmov`, so the payload
read that means `Err` lives in a block that can still reach both variants. A backward walk over
def-use closes it: an op **every** one of whose uses is in a variant-`k` context is a variant-`k`
op, where a `MULTIEQUAL` input slot's context is its own **in-edge**'s variant set rather than the
phi block's. Reading phi inputs as edges is also what makes def-use acyclic, so the walk is exact
rather than a fixpoint guess. A single use whose own context is still ambiguous **BLOCKS** the
conclusion rather than merely failing to narrow it — the difference between a proof and a
majority vote. (That strictness costs a real label; see *What I could not make sound*.)

## What I carried over from #333, and what I left behind

Carried: the *idea* of `variant_regions` + `pin_facets` — a per-block variant set and a pin into
both resolution caches. Everything is re-keyed on the **DWARF discriminant**, and #333's
2-variant `u8` bitmask hack is generalised to an N-variant bitset (the shipped fixture's
3-variant `Three` and the witness set's 3-variant `Trio` are both exercised).

**Left behind, deliberately:** `analyze()`, the classify-by-shape premise, `synthesize()`,
`resolve_name()`, and the `op_return_variant_ctor_ir` printer seam — the one measured to emit a
side-effecting call twice.

## The seam I had to change, and why

The brief proposed gating the printer on `ResolvedUnion::is_locked()`. **I verified that and it
does not hold**, so I built a dedicated proof channel instead and am reporting the substitution.

`ResolvedUnion::update` copies the field number and the data-type but **never the lock**, and the
two paths that rebuild a resolution onto a *new* op — `resolve_in_flow`'s address-based
materialization and `ActionSetCasts::resolveUnion`'s spliced `PTRSUB` — both construct a fresh
unlocked `ResolvedUnion`. Those are precisely the ops the printer descends with: in `use16` the
`Err` payload is read at a `CPUI_CAST` created *after* this pass runs. Gating on the lock would
have named nothing.

So the pin still goes to both caches, locked (which is what stops `ScoreUnionFields` from
overwriting the field), and the **proof** is a separate `Funcdata` side channel keyed by op time
*and* by instruction address. The coarse key answers only for an op created **after** the
analysis ran (`kuna_seal_variant_proofs` records the horizon), and an address carrying two
different proved members is dropped rather than resolved by luck.

**There is a third printer seam, and it is the one that mattered.** The two the brief named
(`push_partial_symbol_ir` at 6631, `op_ptrsub_ir`) are not what a register-returned Rust enum
reaches: its payload half is a `SUBPIECE` of the returned pair whose type IS the overlay union,
so it renders through `push_implied_field_ir` — a function whose doc comment claimed it was
inert. All three are now covered by one helper (`PrintC::variant_member_name`).

### 1. The acceptance bar: the #333 adversarial witness set

Rebuilt in full, **with `-C debuginfo=2` / `-g`** so the DWARF gate is genuinely exercised rather
than trivially empty. `decompile-all --json`, off vs on, keyed by address:

| witness (source truth) | result |
|---|---|
| `#[repr(C)] struct Pair{kind:u64,val:u64}` | **unchanged** |
| `(u64,u64)` tuple, first elem 7 / 9 | **unchanged** |
| `(u64,u64)` bitmask pair | **unchanged** |
| `&'static str` — a plain fat pointer (its "tag" is the `.rodata` address `0x2001ec`) | **unchanged** |
| `Option<&'static str>` — niche-encoded | **unchanged** |
| 3-variant enum returning only two of its tags | **unchanged** |
| C `struct{u64 code; u64 val;}` — gcc `-O2` `.o` linked into the rustc image | **unchanged** |
| C `struct{u64 code; u64 val;}` — clang `-O2` `.o` linked into the rustc image | **unchanged** |
| `fn(u32) -> Result<u64,u64>` consumer (the true positive) | **changed ✅** |

**1 of 14 functions changes on each image, and it is the true positive.** The same C witnesses
compiled standalone (`gcc -g -O2`, `clang -g -O2`, plus a hand-rolled
`struct { enum kind; union value; }` tagged union): **0 of 15 changed, both compilers.**

The gate that makes this structural is not the DWARF gate alone — it is that **only a layout with
a SUPPRESSED payload facet is considered at all**. An enum `dwarfvariants` already names is
skipped before any analysis runs, so this pass provably cannot overwrite a label the layout
already forces. `Option<&'static str>` above is unchanged for that reason, not by luck.

### 2. The refusals are not vacuous

* **`p_either`** — `match r { Ok(v) => v, Err(e) => e }`, both arms identical, so rustc reads the
  payload **once with no guard above it**. It renders `p_mk(x).payload.field_0x8` in both arms.
  This is the "a block that can still reach two variants must stay UNPINNED" witness, and it is
  built to be one.
* **`put_res` / `r16`** (committed fixtures) — rustc computes a `Result`'s discriminant
  branchlessly (`dst->tag = (x < 0xb)`), so no literal ever reaches the tag. Both keep the offset
  spelling, **identical off and on**.
* **`put_opt` / `ret_three`** — already named by the layout, so the layout is never considered.
  Identical off and on.

### 3. The positive cases, hand-checked against source

**`use16` — I checked the arms by hand first, because getting `Ok` and `Err` swapped is the exact
failure DIV-87's suppression rule exists to prevent.** `Ok` has `DW_AT_discr_value` 0; the arm
guarded by `!(v2.tag & 1)` is the one that renders `Ok`, and it carries no arithmetic. The `+ 100`
arm — the one rustc hoisted above the branch — renders `Err`. That matches
`match r16(x) { Ok(v) => v, Err(e) => e + 100 }` exactly, and `the_arms_are_not_swapped` in
`verify_variantguard.rs` pins it.

Committed fixtures (`dwarfvariants_x86_64`, `dwarfvariants_overlay_x86_64`) — no new fixture built:

```
use16     OFF: v2.payload.field_0x8 on BOTH arms
          ON : (int8)v2.payload.Err + 100   /   v1 = v2.payload.Ok;
ret_multi OFF: (rethidden->payload).field_0x4.a = x;   (rethidden->payload).field_0x8.field_0x8 = ...
          ON : (rethidden->payload).P.a = x;           (rethidden->payload).Q.field_0x8 = ...
ret_three ON : unchanged (A/B already forced by disjoint ranges)
ret_niche ON : unchanged (Some already forced)
list_len  ON : unchanged (Cons already forced)
put_res   ON : unchanged (branchless discriminant)
```

### 4. Breadth on a real `-g` Rust binary

A std-linked `rustc 1.90 -C opt-level=1 -C debuginfo=2` witness using `HashMap`, `BTreeMap`,
`Box<dyn Error>`, `Result<u64,u64>`, `Result<u64,u32>`, `Result<Pair,String>`,
`enum Tree { Leaf(i64), Node(Box<Tree>,Box<Tree>) }` and
`enum Expr { Lit, Bin, Neg }` — **161 variant parts, 608 functions**.

**4 functions change. 11 variant labels appear** (revision 3's kill discipline cost 1
function and 6 labels, all of them reads below a call — see above). Every one hand-checked against the source, zero
mislabels.** If it is 5 functions, it is 5.

The important part is *which* five. DIV-87 measured the pre-suppression, shape-scored labels on
this same construct and found them **reproducibly wrong** — "`depth` labelled the `Node` right
child `Leaf.__0`; `eval` labelled the `Neg` payload `Bin.__0`". Those are the two the guard now
gets right:

```diff
  # eval   —  match e { Expr::Lit(v) => *v, Expr::Bin(a,b) => eval(a)+eval(b), Expr::Neg(a) => -eval(a) }
    while (*e).tag == 1 {
-     v1 = eval((*e).payload.field_0x8.__0);   e = (*e).payload.field_0x8.__1;
+     v1 = eval((*e).payload.Bin.__0);         e = (*e).payload.Bin.__1;
    if (*e).tag != 0 {
-     e_00 = (*e).payload.field_0x8.__0;  v1 = -eval(e_00);
+     e_00 = (*e).payload.Neg.__0;        v1 = -eval(e_00);
    } else {
-     v1 = (*e).payload.field_0x8.__0;
+     v1 = (*e).payload.Lit.__0;

  # depth  —  match t { Tree::Leaf(v) => *v, Tree::Node(l,r) => 1 + depth(l).max(depth(r)) }
-     v2 = depth((*t).field_0x0.field_0x8);   ...   v1 = (*t).field_0x8.field_0x8;
+     v2 = depth((*t).Node.field_0x8);        ...   v1 = (*t).Leaf.field_0x8;
```

`Tree` is niche-encoded (`Leaf` = the null box, `Node` = the DEFAULT variant), so `depth` is also
the case that exercises the exact default-variant counting: `tag == 0` admits `Leaf` alone and
`tag != 0` admits `Node` alone. Plus `use_divide` (`Ok`/`Err` as in `use16`) and both
`drop_in_place` glue functions.

### 5. C collateral — `decompile-all --json`, off vs on

| target | changed / total |
|---|---|
| `tests/bug-repro/grep` | 0 / 449 |
| `tests/bug-repro/sort` | 0 / 490 |
| `tests/bug-repro/faillog` | 0 / 69 |
| `tests/bug-repro/libselinux.so.1` | 0 / 621 |
| `tests/bug-repro/betaflight_STM32F405.elf` (ARM Cortex-M) | 0 / 5797 |
| `/usr/bin/ncat` (OpenSSL) | 0 / 416 |
| `/usr/bin/htop` | 0 / 620 |

**0 changed / 8,462 functions, 201,481 lines.** This one is nearly vacuous on its own (no DWARF
variant parts ⇒ empty side table ⇒ the pass returns after one map read), which is why the
`-g`-built adversarial set in §1 is the column that carries information.

### 6. Speed

Interleaved, alternating arms, **pinned to one core**, with the per-function watchdog disabled
(`--max-fn-seconds 0`, so a timeout cannot masquerade as a time), min-of-4:

```
DWARF-heavy witness (161 variant parts, 608 fns)
  pair 1  off 161.18  on 161.14
  pair 2  off 161.05  on 161.07
  pair 3  off 160.94  on 161.06
  pair 4  off 160.72  on 161.62
  min-of-4: 160.72 -> 161.06   +0.21%

null control `grep` (C, zero variant parts)
  pair 1  off 13.65  on 13.64
  pair 2  off 13.65  on 13.65
  pair 3  off 13.66  on 13.68
  pair 4  off 13.68  on 13.68
  min-of-4: 13.65 -> 13.64     -0.07%
```

The control's eight runs span 13.64–13.68 s, which is what makes it credible as "free":
with no variant layouts the pass is one empty-map read per function. Budget is 5%.

## Why default-ON

It is inert without DWARF variant layouts (§5, and structurally: the side table is empty), it
cannot touch a label `dwarfvariants` already forces (§1), it is 0/675 on the datatest corpus by
construction, and it is inside the speed budget. The whole point of DIV-87's suppression was
"name it when we can prove it" — the proof is now here, and leaving it off would ship the proof
and keep the silence.

## What I could not make sound, stated plainly

* **A producer store names no READ at all**, not even one below it — see revision 2. Doing
  that soundly needs a clobber analysis over aliasing (a call handed the pointer may store
  through it), and a producer store does not reach out of its own block either, so a
  construction split across blocks recovers nothing.
* **A read consumed BOTH by the discriminant test and inside one arm is refused.** On a **niche**
  encoding those are literally the same bytes — `Tree` stores the tag in `Node`'s first box — so
  the `Tree` drop glue names the arm's read (`(*a0).Node.field_0x8`) and leaves the guard's read
  on the offset spelling (`(*a0).field_0x0.__0`). A looser rule (intersect over uses, letting an
  ambiguous use fail to constrain rather than block) *does* name it, and I had that working before
  I talked myself out of it: on an execution where the other variant is live, that read still
  happens and its result is still used, so the label would be a claim the other arm contradicts.
  Closing it properly needs a **per-ACCESS width rule** — an 8-byte read at offset 0 of that
  `Tree` can only be `Node`, because `Leaf` claims no byte there — which is a layout question, not
  a guard question, and is a different feature.
* **A branchless producer names nothing.** rustc at `-O1` computes a `Result`'s discriminant as
  `(x < 0xb)`. `put_res` and `r16` are permanently on the offset spelling until something reads a
  *computed* tag, which is a different premise.
* **A switch / jump-table dispatch on the tag is not a seed.** Only equality-shaped conditions
  are read. A `match` over 4+ variants that rustc lowers to a jump table gets nothing.
* **The facet's own interned TYPE name is never substituted**, so a cast still prints
  `Result<u64, u64>::field_0x8` beside a `.Ok`. The type is global to the program; the proof is
  per-access. Undoing that suppression would give an *unpinned* access a scorer-chosen name,
  which is the coin flip DIV-87 exists to prevent.
* **The FIELD inside a named facet keeps `dwarfvariants`'s own per-field suppression** —
  `ret_multi` renders `(rethidden->payload).P.field_0x8`, not `.P.b`. That is a separate rule in
  the importer and this PR does not touch it.
* **The address-keyed proof is at instruction granularity.** It answers only for ops created after
  the analysis ran, requires the field index to match, and drops any address where two members
  were proved — but two distinct objects of the same enum type touched by one machine instruction
  would defeat it. I could not construct such a case on x86-64.

## Tests

* **10 unit tests** (`p5_types/kuna_variantguard/tests.rs`): the scope gate, the union-member ↔
  variant index mapping across a fieldless variant, equality and masked guards, the exact
  default-variant counting (including the ambiguous case that must refuse), an unsatisfiable
  predicate, `population` at every tag width, and the constant-store tag slice.
* **16 cross-crate e2e tests** (`kuna-console/tests/verify_variantguard.rs`), twelve of them
  REFUSALS, and **ablating the option fails 12 of the 16**: `the_arms_are_not_swapped` (checked against the source, first),
  `a_branchless_discriminant_names_nothing`,
  `a_payload_read_that_can_reach_both_variants_stays_unnamed`,
  `an_already_forced_layout_is_never_reconsidered`.
* **22 stage assertions** (`tests/stages/kuna-variantguard.xml`), two-pass off/on, including
  three that must be **identical in both passes**, six added in revision 2 for the
  producer-ordering rule and six in revision 3 for the guard kill.
* `verify_dwarfvariants.rs` now pins `variantguard` **off** — the same discipline it already
  applies to `dwarfstructs` (pinned on), so it keeps measuring the variant-part import alone.
  Two of its assertions previously pinned "nothing may be named" on `use16`/`ret_multi`; those
  are now the *import's* claim, and the shipped rendering is pinned in `verify_variantguard.rs`.

## The three places that said this did not exist

Updated, because they are now wrong: `p0_knowledge/kuna_dwarfvariants.rs` (module header +
`VariantLayout`'s "read by nothing"), `analyzers/dwarf/kuna_dwarfvariants.rs` (the "not attempted
here" paragraph), and `docs/spec/01-program-prep.md` (the closing paragraph of the
`dwarfvariants` bullet).

## Gates (verbatim)

```
######## make test
exit: 0

=== baseline parity ===
PARITY OK
######## make test-stages
datatests: 590/590 assertions passed
exit: 0

=== baseline parity ===
PARITY OK
######## make rust-test
MAKE_EXIT=0
######## make check-spec
python3 tools/check_spec.py
check-spec OK (lenient mode)
######## check_spec --strict
check-spec OK (strict mode)
######## kuna catalog --check
```

`make rust-test`: **330 suites, 4,924 passed, 0 failed**, 39 ignored, 0 failure blocks, 0 targets
failed. (Internal-branch PRs skip the workspace suite in CI, so this was run locally and is the
gate.)

`docs/baseline.json` is **untouched**. `docs/baseline-stages.json` gains only the 22 new
`VARIANTGUARD #1`..`#22` rows and the 568 → 590 footer; nothing was removed.

**Counts re-derived from the actual failing assertions:** settables 124 → **125**; tiers
(26, 51, 47) → (26, **52**, 47); catalog JSON row commas 123 → **124**; `with_live` 36 → **37**;
`catalog_bytecompat.rs` fixture 124 → **125** (fn renamed `fixture_has_all_125_settables`);
universal-tree leaves 274 → **275**; stages corpus files 220 → **221**.
`tests/fixtures/list_action_decompile_oracle.txt` and
`tests/fixtures/phase_catalog.json` regenerated.

https://claude.ai/code/session_011KpG7qK6BDFZyPnyo4r1c1


mahaloz:
[AUTOMATED] **Held as a proposal — the naming does not ship.** Recording the decision and the arc, because the arc is the deliverable.

## Why

The premise is sound and is *not* #333's: #334 already recorded the compiler's own answer (`DW_AT_discr` + per-variant `DW_AT_discr_value`), so only the block→value mapping is derived. Nothing here infers a type from codegen shape.

The problem is the mapping. **Four independent adversarial rounds each found a different shape of one defect — a fact about an enum object's discriminant applied at a program point where it no longer holds.** Every round shipped green gates and non-vacuous tests, and every round was wrong:

| round | defect | witness |
|---|---|---|
| R1 | a producer `tag = K` store propagated backwards over whole blocks, naming a read **above** it | `read_then_clobber` / `read_then_clobber_ok` — byte-identical but for the clobber constant, giving the **same instruction opposite names** |
| R2 | a guard fact had no kill at all, so a read or write after a store or a call still carried it; a singleton guard region also discarded a correct producer fact | `guarded_write` (`tag = 1;` then `.Ok.__0 = v1 + 1;`), `read_after_call_{ok,err}_guard` |
| R3 | a guard's `CBRANCH` *position* is last in its block, but its *condition Varnode* is not — rustc hoists the tag read above the clobber, the block's fact is correctly raised to ⊤, and the same block's stale edge constraint resurrects it in a successor the positional kill never reaches | `hoist_read`, `hoist_xblock` |

The last one, verbatim at the shipped default, with kuna's own rendering of `wipe` two lines away being `(*dst).tag = 1; (*dst).payload.Err.__0 = 99;`:

```
v1 = (*dst).tag;
wipe(dst);
if v1 & 1 != 0 { return 0; }
return (*dst).payload.Ok.__0;     // the object is Err(99)
```

A wrong variant name is a confident lie about which arm of a `match` the code is in — strictly worse than the `field_0x8` DIV-87 deliberately prints. Rather than attempt a fourth positional patch, the conclusion is recorded: **the memory-object guard is not soundly reachable by positional/CFG reasoning.** It needs reaching-definitions plus an escape discipline over the object, which is separate work.

## What is verified sound, and is the retry's starting point

- **The producer half** — writes only, intra-block, positional, killed by a tag store / any call / any unattributable store — survived every attack in rounds 2 and 3.
- **The value-object guard** held under both reviewers' probes: a `MULTIEQUAL` joining two variants, a value round-tripped through memory, and an address-taken local handed to a callee (correctly reclassified as memory, so the carve-out did not leak). An SSA value cannot be clobbered, which is exactly why the resurrection has no analogue there.

**Recommended retry: value-object guard + producer writes, memory-object guard omitted until there is an escape analysis.** That configuration has not itself been adversarially tested and would need its own round.

## Cost, measured

On a real `-g` Rust binary, dropping the memory guard costs 5 functions/17 labels → 4/11 on one witness and 5/9 → 3/7 on an independent reviewer's; both agree in kind — every lost label is a read below a call, concentrated in drop glue and recursive walks.

## Two things worth more than the feature

1. **The gates are structurally blind here.** Both parity corpora are symbol-less bytechunks with no DWARF, so nothing in `make test` / `make test-stages` can see this pass at all. Every defect above was found only by a purpose-built fixture written against a specific claimed invariant.
2. **Non-vacuous tests were necessary and nowhere near sufficient.** Ablation showed 12 of 16 cross-crate tests and 15 of 22 stage assertions genuinely fail with the pass off — and the code was still wrong three times over.

A reported CLI-vs-console masking effect **did not reproduce** and is withdrawn: `kuna decompile` renders a Rust binary *as Rust* (`(*dst).payload.Ok`) while the console prints C (`(dst->payload).Ok`). Same name, different output language — easy to misread when grepping for one spelling.

## State

Option **default-OFF** and in `EXCLUDED_ON_PURPOSE`, so no preset reaches it; its catalog row opens with the known-unsound statement and names the witnesses. `docs/history.md` is byte-identical to `main` — DIV-97 removed, since nothing diverges. The implementation stays on the branch as the reference implementation. `docs/features/variantguard/analysis.md` carries the full account, and all 15 witness sources are committed under `witnesses/` with the defect each caught and its build line, so the next attempt starts from evidence instead of re-deriving it.

Gates green with the option off: 675/675 and 590/590 PARITY OK, 330 suites / 4,924 passed / 0 failed, check-spec strict OK, catalog OK.

PR #352, opened 2026-08-28
fix(ghidra): apply the host's typelocked parameter types (wrong flag namespace)

psifertex:
## The bug

`RemoteScope::materialize` built every `ParameterPieces.flags` from the **`varnode_flags`** namespace instead of **`parameter_pieces_flags`**:

| constant | value | |
|---|---|---|
| `varnode_flags::typelock` / `namelock` | `0x100` / `0x200` | what was written |
| `parameter_pieces_flags::TYPELOCK` / `NAMELOCK` | `0x10` / `0x08` | what is read |

`0x300 & 0x10 == 0`, so every parameter materialised from Ghidra's `<localdb>` was stored type- and name-unlocked no matter what the host sent. This was the only site in the workspace constructing a `ParameterPieces.flags` from `varnode_flags`; the analogous console `map param` path (`kuna-console/src/ifacedecomp.rs`) already used the correct namespace.

## Causal chain

1. `apply_locked_prototype_with_model` seeds the prototype correctly, then `apply_mapped_params` overwrites it with the mis-flagged pieces (`infra/decompile_drive.rs`).
2. `ProtoStoreInternal::set_input` stores `pieces.flags` verbatim (`p4_calls/fspec.rs`).
3. `FuncProto::is_input_locked()` masks with `parameter_pieces_flags::TYPELOCK` → false.
4. `ActionPrototypeTypes` is therefore skipped, so `update_type_locked` never forces the declared type onto the input varnode (`p4_calls/coreaction_protos.rs`), and `clear_unlocked_input` drops the host's names, leaving `kuna_arg_name` to re-derive `a0`/`a1`.

**The wire decode was never at fault.** `decode_symbol_header` reads `ATTRIB_TYPELOCK` off the `<symbol>` child exactly where Ghidra's `HighSymbol.encodeHeader` writes it, and `RemoteProto::is_input_locked()` is true off the wire.

## Evidence

Ghidra 12.1.2, aarch64 ELF with DWARF, `main`. Measured on the Ghidra side via `LocalSymbolMap.grabFromFunction`: `argv` = `char **`, storage `x1:8`, `typeLocked=true`, `nameLocked=true`. A tee shim on the decompiler pipe confirms the packed mapped-symbol record reaching `kuna_ghidra` is byte-identical to the one the stock C++ `decompile` receives — and stock renders `argv[1]` correctly from it.

```
before:  int main(int a0,long a1)
           int argc;   // w0
           long argv;  // x1
         file = *(char **)(argv + (long)argc * 8 + -8);

after:   int main(int argc,char **argv)
         file = argv[(long)argc + -1];
```

With `argv` typed, the machine's `* 8` is absorbed by the pointee size and all 27 `argv` accesses in that function recover array notation. No structural transform changed: branchflip 54, no-return 29, ternary 1, before and after.

Retyping from the GUI could not work around this — neither `Parameter.setDataType(..., USER_DEFINED)` nor `HighFunctionDBUtil.updateDBVariable` (the Decompiler window's *Retype Variable*) changed the output, because the type was already arriving correctly and being discarded after decode.

## Scope

Not register-specific, despite how it presents. The locals path (`seed_mapped_symbols`) legitimately masks with `varnode_flags` — `Symbol` attributes really do live in that namespace — and only the params path crosses into `ParameterPieces`. aarch64 makes it look like a register-parameter bug because all of `main`'s params are register-mapped; a stack-passed parameter (arg 7+ on aarch64, any x86-32 `__cdecl` param) was broken identically.

## Gates

| Gate | Result |
|---|---|
| `make test` | **PARITY OK** — 675/675 |
| `make test-stages` | **PARITY OK** — 568/568 |
| `make check-spec` | green |
| `make rust-test` | green except one pre-existing failure (below) |

`kuna-cli::decompile_project_cli::header_syntax_checks_with_cc` fails **identically on an unmodified tree at the base commit** (`3f98c70`): the generated `fauxware.h` declares `void main(void);` and clang rejects it with `'main' must return 'int'`. Unrelated to this change and probably deserves its own issue. Happy to add the `full-ci` label if you want CI's verdict on `rust-test` on the record.

No option gate, no `docs/spec` change, no baseline re-pin: this is a strict bug fix that only corrects wrong output, and both parity corpora passed unchanged.

## Not included: a regression test

There isn't one, deliberately rather than by omission. The right test targets `RemoteScope::materialize` — feed a typelocked register param, assert the materialised `ParameterPieces.flags` carries `parameter_pieces_flags::TYPELOCK` — which fails before this change and passes after. Say the word and I'll add it as a second commit.

### DIFF
--- a/decompiler/crates/kuna-decomp/src/infra/remote_provider.rs
+++ b/decompiler/crates/kuna-decomp/src/infra/remote_provider.rs
@@ -1411,8 +1411,8 @@ impl RemoteScope {
                                     crate::fspec::ParameterPieces {
                                         addr: rp.storage.clone(),
                                         type_: Some(dt),
-                                        flags: varnode_flags::typelock
-                                            | varnode_flags::namelock,
+                                        flags: crate::fspec::parameter_pieces_flags::TYPELOCK
+                                            | crate::fspec::parameter_pieces_flags::NAMELOCK,
                                     },
                                 ));
                             }
--- a/decompiler/crates/kuna-decomp/src/infra/remote_provider/tests.rs
+++ b/decompiler/crates/kuna-decomp/src/infra/remote_provider/tests.rs
@@ -387,10 +387,9 @@ const CAPTURED_MAIN_MAPSYM: &str = r#"<doc id="0x0"><mapsym>
   <rangelist/>
 </mapsym></doc>"#;
 
-/// Decode the vendored REAL Java answer (XML form; the packed wire form
-/// differs only in the marshaling layer behind `dyn Decoder`).
-#[test]
-fn captured_java_function_mapsym_decodes() {
+/// The capture's parameters carry `register` storage, so its manager needs the
+/// space the plain [`manager`] omits.
+fn manager_with_registers() -> Rc<AddrSpaceManager> {
     let mut m = AddrSpaceManager::new();
     m.insert_space(Rc::new(AddrSpace::new(
         spacetype::IPTR_PROCESSOR,
@@ -416,9 +415,13 @@ fn captured_java_function_mapsym_decodes() {
         0,
     )))
     .unwrap();
-    let m = Rc::new(m);
-    let types = types_with_core();
+    Rc::new(m)
+}
 
+fn decode_captured_main(
+    m: &Rc<AddrSpaceManager>,
+    types: &TypeFactoryImpl,
+) -> Box<RemoteSymbolRecord> {
     let mut registry = IdRegistry::with_base_ids();
     register_remote_provider_ids(&mut registry);
     crate::dtype::register_type_wire_ids(&mut registry);
@@ -428,11 +431,20 @@ fn captured_java_function_mapsym_decodes() {
 
     let store = kuna_base::xml::xml_tree(CAPTURED_MAIN_MAPSYM.as_bytes()).unwrap();
     let root = store.get_root().clone();
-    let mut dec = XmlDecode::new_with_root(&m, &registry, &root, 0);
-    let rec = match decode_mapped_answer(&mut dec, &types).unwrap() {
+    let mut dec = XmlDecode::new_with_root(m, &registry, &root, 0);
+    match decode_mapped_answer(&mut dec, types).unwrap() {
         RemoteMapAnswer::Symbol(rec) => rec,
         _ => panic!("expected a symbol"),
-    };
+    }
+}
+
+/// Decode the vendored REAL Java answer (XML form; the packed wire form
+/// differs only in the marshaling layer behind `dyn Decoder`).
+#[test]
+fn captured_java_function_mapsym_decodes() {
+    let m = manager_with_registers();
+    let types = types_with_core();
+    let rec = decode_captured_main(&m, &types);
     assert_eq!(rec.symbol_id, 0x18b);
     let func = rec.func.as_ref().expect("function record");
     assert_eq!(func.name, "main");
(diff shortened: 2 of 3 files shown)


psifertex:
Claude argued with me to drop the `[automated]` prefix since I was involved in the testing / validation / approval but... `¯\_(ツ)_/¯ `

mahaloz:
Reviewed in depth. **The diagnosis is right and the chosen value is right** — merging once CI is green. I pushed one commit to the branch (test + spec); details and two follow-ups below.

## Confirmations

Upstream independently corroborates the diagnosis: `ProtoStoreSymbol::setInput` (`fspec.cc:3155-3182`) builds an explicit `mirror` word translating `ParameterPieces::typelock` → `Varnode::typelock`, and `fspec.hh:360-368` annotates each pieces bit *"mirrors Varnode::…"* — the two namespaces are deliberately non-interchangeable. `TYPELOCK|NAMELOCK` is also exactly what upstream uses for this shape (`ifacedecomp.cc:625`), matching our own sibling builder at `kuna-console/src/ifacedecomp.rs:1198`. A mechanical sweep of all 8 non-test `ParameterPieces.flags` sites confirms this was the only crossing, in either direction.

Two refinements to the causal chain in the description, both cosmetic:

- `ActionPrototypeTypes` is **not** skipped — only its locked-input branch (`coreaction_protos.rs:247-296`) is. That is why the "before" output still had the correct `int` return: the output lock lives on the store's outparam, which `apply_mapped_params` never touches.
- The failure is a **clobber**, not a missing application. `set_pieces` had already locked the params (`fspec.rs:5531`); `apply_mapped_params` silently *un*-locked them by replacing `inparam[i]` wholesale, and since `is_input_locked()` reads slot 0 only (`fspec.rs:5132`), one bad slot unlocked the whole signature. Pre-PR, sending the host's storage made output strictly **worse** than sending none.

The `<addr>`-storage premise is confirmed against Ghidra's Java: params always go through `newMappedSymbol` → `MappedEntry`, never the hash branch (`LocalSymbolMap.java:442-455`).

## What I added (eb63eaa2)

Your instinct to offer a test was right, and the bug's invisibility is the reason. `RemoteScope::materialize` is reachable only via `install_remote_provider`, whose sole caller is the Ghidra bridge — so no datatest or stage XML can execute it, and **the ghidra-sim oracle serves cat-0 params with `<hash>` storage** (`ghidra_sim/oracle.rs:537-541`), which `materialize` skips, leaving `param_storage` empty in every simulated session. All four gates were structurally blind; the 0/675 result was guaranteed, not evidence.

Rather than hand-build a wire doc, the test drives the **real Ghidra 12.1.2 capture already vendored** in that file (`CAPTURED_MAIN_MAPSYM`) through `materialize` and asserts the resulting `param_storage` flags. Fails on the old constant with `argc: locks must be parameter_pieces_flags bits, got 0x300`. Also a spec sentence: the owning chapter described the storage echo only as the `checkFullCommit` mechanism and never said the pieces carry lock bits — the exact gap this fell into.

No `docs/history.md` row: the Convergences table is for *"a corpus-wide output shift… for anyone bisecting"*, and this shifts 0/675 and 0/568.

## Gates (local, at eb63eaa2)

`make test` 675/675 · `make test-stages` 568/568 · `make check-spec` OK (lenient **and** strict) · `kuna catalog --check` OK · `make rust-test` **fully green, 329 binaries, 0 failures** · `cargo test -p kuna-ghidra --release --include-ignored` (the ghidra-sim differential harness) green.

On `header_syntax_checks_with_cc`: it **passes** here and on `ubuntu-latest`. It is compiler-dependent — `decompile-project` emits `void main(void);` into the header, gcc accepts it silently, clang errors `'main' must return 'int'`. So your `cc` is clang. Real bug, filed separately; nothing to do with this PR.

## Follow-ups filed, not blocking

Locking these prototypes for real exposes a **pre-existing** defect: Ghidra sends `__return_storage_ptr__` as a cat-0 param with `hiddenretparm="true"`, and `remote_provider.rs:601` explicitly discards that attribute — so `to_pieces` declares it as a real parameter and `update_all_types` synthesizes a *second* hidden-return slot. For a by-value-aggregate return that now prints a duplicate-named trailing parameter.

I checked whether this blocks, and it does not: the echoed `<localdb>` that `checkFullCommit` actually reads is **byte-identical** before and after — 4 cat-0 slots with the phantom `beta`@RCX either way. The corrupt wire contract is already on main; this PR only makes it visible in the printed C. A control run with the storage overrides removed reproduces it identically with your change inert. Measured prevalence: 338/210,380 functions (0.161%) across the decbench corpus return >16 bytes.

Filed with a validated patch. Two notes for whoever takes it: guarding on a count mismatch is *worse* than doing nothing (it yields `(BigStruct *rethidden, BigStruct *__return_storage_ptr__, …)`), and the fix only produces correct output **once this PR is in** — #352 is its prerequisite.

Thanks for the careful write-up — the wire-vs-decode elimination and the "retyping from the GUI can't work around this" note saved real time.

mahaloz:
Mannnn. Why did it not follow my `[AUTOMATED]` rule for my PR comment.

mahaloz:
I actually agree @psifertex, since you were involved and this was not fully blind, I would say it is not normal `[AUTOMATED]`

PR #356, opened 2026-08-29
[AUTOMATED] feat(p9): voidtailreturn — a void function's body ends where the source's does (DIV-98)

mahaloz:
> **Draft — measurement in flight.** The headline number below was measured on the
> feature *before* its fifth guard (commit `831f6782`) was added. A re-measurement
> against the shipped code is running; I will post it before marking this ready.
> **Not self-merged**: this is GED work, outside the type-score mandate I was given.

## What

kuna (faithfully to Ghidra) emits the function's final `CPUI_RETURN` unconditionally, so a void function ends `... }` / `return;` / `}`. The C it was compiled from has no `return` there at all — it falls off the end of the body.

The cost is not only stylistic. pyjoern merges `FUNCTION_END` into its predecessor only when that predecessor's sole successor is `FUNCTION_END`, so a source function whose tail is an `if` or a loop has **no node there**, while kuna's printed statement re-materialises one: +1 node, +1–2 edges, and an `is_exitpoint` role flip that on its own defeats GED's isomorphism test. 68 of the 250 decbench sample-set outputs carry exactly one bare trailing `return;`.

## How

`PrintC::emit_function_body` computes at most one elidable op per function (`elidable_void_tail_return`); `emit_basic_block_ops` skips exactly that op. Five conditions must all hold, and **each has a named counterexample that forced it**:

1. **The prototype returns void** — a `return <value>;` is never redundant.
2. **The op is the tail of the LAST structured leaf**, reached by descending only the containers that print no construct of their own (`Graph`/`Ls`), so a return nested inside an `if`/loop/switch arm is never touched.
3. **That leaf is not an unstructured goto target** — bash `rl_echo_signal_char` prints `label_115e79:` directly above its trailing `return;`; eliding it would leave a label with nothing after it, which is invalid C.
4. **Exactly one structured leaf carries that RETURN op** — `returndup` and `taildup` clone a shared epilogue by *aliasing* one op across several leaves, so suppressing by identity would also delete genuine mid-body early returns. A unique owner makes the elision positional rather than identity-based.
5. **The function prints at least one other statement** — a body whose only statement is the return came back completely empty, which is not the source shape and which strands the ghidra markup with no op to cross-link to the `<ast>`. `kuna-ghidra`'s `decompile_at_e2e` pins that invariant and caught it (`831f6782`).

A `BlockCopy` mirror is resolved through to the bblocks block it stands for (`structured_leaf_tail`) — the normal shape of a structured function's trailing block, and the reason the direct `sblocks_basic_tail` lookup is not sufficient.

## Delivery

The shipped catalog default stays **OFF** (the XML datatest corpus pins the upstream rendering and applies no mode); `AGGRESSIVE_OVERRIDES` carries it, which is what makes the elision the default output for every real binary — exactly the shape DIV-75 uses for `ctypes`. `auto` selects `aggressive` under 500 KiB, so this is the rendering `kuna decompile`, `decompile-all`, `decompile-project`, the web front-end and the benchmark all get.

## Measurement (pre-guard-5; re-run in flight)

`scripts.decbench.optsweep`, bidirectional, whole O0 decbench slice — 32,329 functions / 265 slices, both arms one kuna binary invoked exactly as decbench's `kuna_raw` backend does:

| | base | variant | delta |
|---|---:|---:|---|
| GED-perfect | 15,308 | 15,933 | **+625** |
| aggregate GED | 259,329 | 253,985 | −5,344 |

- **to perfect 687 / off perfect 62** — 11.08 : 1, McNemar z = +22.84
- improved 1,561 / worsened 186
- every architecture net-positive (x86-64 +282, ARM +338, PE-x86 +5); **no project net-negative**

**Controls:** 21,478 functions whose bodies the pass never changes score identically (0 differ); 0 slug-only diffs; the baseline arm agrees with the tree's published GED verdict on 31,584/32,329 (97.7% — the residual is kuna-build drift since the tree was scored).

**The 62 regressions** are the one shape the binary cannot distinguish — the *source* itself ends with an explicit `return;` — concentrated in the betaflight (24) and cleanflight (21) register-poke helpers.

**Speed** +0.17% (coreutils/ls O0, interleaved min-of-5, budget 5%).

## Gates

- `make test` — **675/675 PARITY OK** (no re-pin: the catalog default is unchanged)
- `make test-stages` — **570/570 PARITY OK** with the new two-pass `tests/stages/kuna-voidtailreturn.xml`
- `make check-spec` — green
- `make rust-test` — re-running against the guarded code (it was green on the pre-guard build, and `decompile_at_e2e` — the test that forced guard 5 — now passes)

Catalog counts bumped 125 → 126 settables (tier 27→28 core, live-field 37→38, static-form rows 124→125, corpus files 220→221); `phase_catalog.json` fixture regenerated. Docs: DIV-98 in `docs/history.md`, `docs/spec/09-emission.md` §9 brace-form, `docs/options.md` regenerated.

## Context

This is lever **L1** of the DecBench sample-set plan: on that 250-function slice it is the single change that takes kuna from 75 to 83 and into 3rd place (top deterministic decompiler). Two independent analysis lenses converged on it before it was built.

https://claude.ai/code/session_01FvpHGrvHyKDNmLSPVnZw2G


mahaloz:
## Re-measurement landed — and it decomposes the number

Two sweeps, both complete, landing on the **identical variant value**:

| sweep | binary | option in preset? | base | variant | delta |
|---|---|---|---:|---:|---|
| A | `69343162` | no | 15,308 | **15,933** | +625 |
| B | `831f6782` | **yes** | 15,747 | **15,933** | +186 |

Two things fall out, and they are the reason the draft was held:

**1. Guard 5 costs nothing.** Sweep B ran the guarded build (`831f6782`, the
empty-body condition `decompile_at_e2e` forced). It reaches exactly the same
15,933 perfect functions as the pre-guard build. The guard removes only cases that
were never wins.

**2. Preset membership delivers +439, not +625.** Sweep B's baseline arm already
carries the elision on every binary `auto` resolves to `aggressive`, so
15,747 − 15,308 = **+439 is what this PR actually buys**. The residual **+186 sits
on binaries ≥ 500 KiB**, where `auto_mode_for_size` selects `reliable` — whose
override list is empty. Reaching those needs a catalog-default flip or a
`reliable` override list, and this PR claims neither.

DIV-98 and the catalog `use_when` now state both numbers and which one this change
buys (`977bc669`).

## Gates, final

- `make test` — 675/675 PARITY OK (no re-pin)
- `make test-stages` — 570/570 PARITY OK
- `make rust-test` — **green on the guarded build** (330 result blocks, 0 failures)
- `make check-spec` — green
- `kuna catalog --check` — green

Sweep B's controls: 28,554 unchanged bodies score identically (0 differ), 0
slug-only diffs, baseline agrees with the tree's published verdict on
31,075/32,329.

Still a draft and still not self-merged — this is GED work, outside the type-score
mandate for this session.

PR #401, opened 2026-09-05
feature: json graph export (kuna decompile-graph)

dobin:
For https://github.com/dobin/REvAId I'd like kuna to hand me the whole program as
one JSON document: every function, with its asm and decompiled C, and the call
edges between them. `kuna decompile-all --json` and `kuna xrefs` each answer half
of that, and stitching them together outside kuna means two loads and a
containment model of my own. `kuna decompile-graph` writes it in one in-process
load.

```console
$ kuna decompile-graph decompiler/crates/kuna-analysis/tests/fixtures/fauxware -o g.json --label revaid-0.4

$ jq -c '.functions[] | select(.name=="authenticate") | {name,kind,size,signature}' g.json
{"name":"authenticate","kind":"normal","size":137,"signature":"unsigned long authenticate(char *a0,char *a1)"}

$ jq -c '.edges[] | select(.callerAddress==4195940)' g.json
{"callerAddress":4195940,"calleeAddress":4195664,"kind":"call","calleeOrder":0}
{"callerAddress":4195940,"calleeAddress":4195680,"kind":"call","calleeOrder":1}
{"callerAddress":4195940,"calleeAddress":4195632,"kind":"call","calleeOrder":2}

$ jq -c '.edges[] | select(.calleeAddress==4196125)' g.json   # who reaches main
{"callerAddress":4195712,"calleeAddress":4196125,"kind":"data","calleeOrder":2}
```

That last one is why edges carry a kind: nothing calls `main` on a glibc ELF.
`_start` materializes its address and hands it to `__libc_start_main`, so an edge
list built out of call and branch references alone leaves the one function you
opened the document for with no caller at all.

Compared against a Ghidra export of the same binary
([test-data.zip](https://github.com/user-attachments/files/31860772/test-data.zip)):

```
Function entry addresses
  common:       525
  Ghidra only:  95
  Kuna only:    169
  Kuna recall vs Ghidra:    84.7%
  Kuna precision vs Ghidra: 75.6%

Call-graph edges: (callerAddress, calleeAddress)
  common:       1045
  Ghidra only:  289
  Kuna only:    102
  Kuna recall vs Ghidra:    78.3%
  Kuna precision vs Ghidra: 91.1%
```

Those edge numbers predate the resolution rule below, which drops targets that
are not functions and keeps functions that are only pointed at.

## The command

- Every discovered function is a node. `--functions` / `--addr` narrow which
  nodes get a decompiled *body*, not which appear, so `--functions main` buys the
  whole call graph plus one body. The rest of the flags are `decompile-all`'s,
  through its parser.
- Bodies, kinds, edges and listings each come from the surface that already owns
  them: `decompile-all`'s whole-binary target policy, the per-function classifier
  the browser inventory groups by (moved from `kuna-wasm` to `kuna-console` so
  both read one answer), the call-graph model `--reachable-from` walks, and the
  `kuna disassemble` walk. Edges carry that model's `kuna xrefs` kind — `call`,
  `jump`, `data` — so the document and that command cannot describe one program
  differently, and both ends of every edge are rows of the same document.
- The target policy holds even against an explicit selection: `--addr` onto a PE
  import pointer slot leaves an `import` row with no body rather than lifting a
  plausible-looking function out of a pointer table, and the run says so on
  stderr. An undecodable byte inside a real body is a `.byte` row rather than a
  dropped listing.
- `address` is the key, not `name`. A PLT thunk, the import slot it forwards
  through and the callable they stand for are three rows under one name.
- The document is C. `codeC` names its language, so `--language rust` is an
  error here and the auto-language policy that follows a rustc-built binary is
  off, exactly as for `decompile-project`.
- A failed function's `error` travels in its row, next to the assembly that is
  left to look at. `kind` does not depend on what the run was asked to
  decompile, and two runs of one command are byte-identical.
- `XrefIndex::has_indirect_calls` stays. A computed call files no edge because it
  has no static target, so without it "this function calls nothing" and "this
  function's callee is decided at run time" are the same answer. It is `CALLIND`
  only — a forwarding veneer's `jmp [slot]` has a knowable destination and is
  reported as `forwardsTo` instead — and the call site is attributed to the
  function containing it, the rule the reference buckets beside it already use.
- Named `decompile-graph` to sit beside `decompile-all` and `decompile-project`.
  The free-text second positional is now `--label`.

Schema and an example document: `docs/cli.md`; the reasoning behind each rule:
spec §9.7.

## Tests

`decompile_graph_cli.rs` covers the document shape and the file export, plus the
rules that would otherwise regress silently: `_start -> main` is an edge, a PE
import slot gets a label and no C even when `--addr` names it, every edge
endpoint is a row of `functions`, a rustc-built binary still exports C, and two
runs produce the same bytes. `verify_indirect_callers.rs` pins what
`has_indirect_calls` means on `fauxware` — a `CALL RAX` is reported, `main`'s
nine direct call sites are not, and neither is a PLT stub's `JMP qword ptr [...]`.

The 896 KB / 1073-function `mcount_x86_64` fixture exports in 78.2 s, against
83.8 s for `decompile-all --json` on the same input (interleaved, best of two).

https://claude.ai/code/session_01TzfafnkDMQiYoifXP5HSC8


mahaloz:
Merged — thanks, this was a real gap.

A few things changed on the way in that REvAId will care about, since you said you hadn't read `graph_export.rs`:

- The command is now `kuna decompile-graph`, and the document is `schemaVersion: 4`.
- **Edges carry a `kind`** (`call` / `jump` / `data`). The old edge set was filtered to call and branch references, which meant `main` had no caller at all on a glibc ELF — `_start` hands its address to `__libc_start_main` as a pointer. Address-taken callees are now `data` edges. Filter on `kind` if you only want call sites.
- **`address` is the key, not `name`.** Names repeat within one document: a PLT thunk and the import slot it forwards through are both `printf`; `pe_imports.exe` carries `LeaveCriticalSection` five times.
- **Bodyless rows are honest now.** `kind` `import`/`data`/`external` rows have `codeC`, `assembly`, `signature` and `error` all `null`. Previously `--addr <import slot>` fabricated a body for a pointer slot.
- `--language rust` is refused here rather than putting Rust in a field called `codeC`; use `kuna decompile` or `decompile-all --json` for other output languages.
- Ordering is deterministic — two runs are byte-identical — so you can diff documents across builds.

`indirect_callers` stayed, but it is exposed as `hasIndirectCalls` per function and its attribution now matches the rule `kuna xrefs --from` uses, rather than whichever descent happened to reach the instruction first.

Full field-by-field reference is in `docs/cli.md` under `kuna decompile-graph`.

Issue #403, opened 2026-09-05
Function not correctly decompiled: "funcboundflow: fall-through reached the next function entry; truncating flow here"


dobin:
My main function doesnt get correctly decompiled. 

https://github.com/dobin/redtest

See the ghidra-asm where the decompilation of kuna stops. 

# Kuna 

Decompiled function which is too short: 

```
// Function: sub_140002650 @ 0x140002650
void sub_140002650(void)
{
  unsigned short v1;
  unsigned long long v10; // stack - 0x20
  unsigned int v2; // eax
  char v3 [8];
  char v4 [96];
  char *v5; // rsp
  unsigned int v6; // stack - 0x30
  unsigned int v7; // stack - 0x2c
  unsigned int v8; // stack - 0x28
  unsigned int v9; // stack - 0x24
  
  v5 = v3;
  v10 = dat_140027000 ^ (unsigned long long)v3;
  sub_140001080("RedTest\n");
  v6 = *(unsigned int *)&dat_140027a88[-8];
  v7 = *(unsigned int *)&dat_140027a88[-6];
  v8 = *(unsigned int *)&dat_140027a88[-4];
  v9 = *(unsigned int *)&dat_140027a88[-2];
  if (dat_140027aa0) {
    (*(void *)swi(0x29))(5);
    v5 = v4;
  }
  dat_140027aa8 = 0;
  dat_140027ab0 = 0;
  *(unsigned long long *)&v5[-8] = 0x1400026c9;
  if (!EventRegister(&v5[0x68],sub_140001000,0x140027a80,0x140027aa0)) {
    v1 = *dat_140027a88;
    *(unsigned long long *)&v5[-8] = 0x1400026ea;
    EventSetInformation(dat_140027aa0,2,dat_140027a88,v1);
  }
  do {
    *(unsigned long long *)&v5[-8] = 0x1400026f1;
    v2 = sub_14000974c(0);
    *(unsigned long long *)&v5[-8] = 0x1400026f8;
    sub_1400094c8(v2);
    *(unsigned long long *)&v5[-8] = 0x140002702;
    dat_140028bc8 = sub_140009bd0();
  } while ; // warn: funcboundflow: fall-through reached the next function entry; truncating flow here
  *(unsigned long long *)&v5[-8] = 0x140002d77;
  sub_140002460("[!] Failed to allocate shellcode buffer");
  *(unsigned long long *)&v5[-8] = 0x140002d81;
  sub_140009490(1);
[...]

PR #410, opened 2026-09-05
[AUTOMATED] fix(p2): funcboundflow's truncation must start a basic block

mahaloz:
## The problem

`funcboundflow` stops a function's flow where the fall-through runs into the
next function's entry. When that fall-through belongs to a *conditional* branch
the truncation lands inside the branch's own block, and the emitted C is not
valid C. Both shapes decompile fine with `--option funcboundflow off`.

```
$ cat repro.s
.text
.globl branchcut ; .type branchcut,@function
branchcut:                    # last insn is a conditional branch,
  mov  %edi,%eax              # whose fall-through is neighbor's entry
  add  $15,%eax
  test %eax,%eax
  je   cold
.globl neighbor ; .type neighbor,@function
neighbor:
  lea  -5(%rdi),%eax
  ret
cold:
  mov  $99,%eax
  ret
.globl loopcut ; .type loopcut,@function
loopcut:
  xor  %eax,%eax
1:add  $15,%eax
  cmp  $100,%eax
  jl   1b                     # fall-through is neighbor2's entry
.globl neighbor2 ; .type neighbor2,@function
neighbor2:
  lea  -7(%rdi),%eax
  ret

$ as -o repro.o repro.s && ld -Ttext=0x100040 -e branchcut -o repro.elf repro.o

$ kuna decompile repro.elf branchcut
unsigned long branchcut(int a0)
{
  do {
  } while ; // warn: funcboundflow: fall-through reached the next function entry; truncating flow here
  return 99;
}

$ kuna decompile repro.elf loopcut
void loopcut(void)
{
  /* WARNING: decompilation failed: decompile pipeline reached an un-ported seam (LOSS-131): index out of bounds: the len is 1 but the index is 1 */
}
error: decompilation failed for loopcut in repro.elf: decompile pipeline reached an un-ported seam (LOSS-131): index out of bounds: the len is 1 but the index is 1
note: decomp_dbg stderr:
thread 'main' panicked at crates/kuna-decomp/src/substrate/block.rs:640:23:
index out of bounds: the len is 1 but the index is 1
```

## The fix

- The artificial no-return RETURN that replaces the truncated fall-through now
  starts a basic block. `collectEdges` emits a fall-through edge for every
  CBRANCH whatever else the walk decided, so without a boundary that edge
  pointed back at the branch's own block: the self-edge the structurer renders
  as `} while ;`. This one change fixes both shapes.
- The halt is deliberately not marked an instruction start, because `fallthruOp`
  can only resolve the fall-through to it as a same-instruction op: the
  truncated address is never decoded.
- `ActionConditionalConst` checks for two out-edges before reading them. In
  `loopcut` the self-edge and the branch's taken edge landed on the same block,
  so the CBRANCH block had one out-edge and `getOut(1)` ran off the end. The
  check turns that into a skipped pass instead of a dead process; it does not
  make the C right on its own, since with the block boundary still missing
  `loopcut` comes out as `do { v1 += 0xf; (v1 < 100); } while( true );`.

Truncation at a non-branch instruction, which already put the halt in its own
block, is unchanged. On a PE crackme `kuna decompile-all` goes from 7
`} while ;` occurrences across 32 truncations to 0 with the truncation count
identical, and binaries with no truncation are byte-for-byte identical.

## The tests

`tests/stages/gh403-funcboundflow-cfg.xml` carries both shapes as a flat
bytechunk with function symbols. Without the fix the `} while ;` assertion and
the "loopcut still decompiles" assertion fail; the two that pin the truncation
itself pass either way. Datatests stay 675/675 and the workspace suite is green.

Fixes the invalid-C half of #403; the `.pdata` function-boundary half is separate.

https://claude.ai/code/session_01TzfafnkDMQiYoifXP5HSC8

### DIFF
--- a/decompiler/crates/kuna-base/src/xml.rs
+++ b/decompiler/crates/kuna-base/src/xml.rs
@@ -1797,7 +1797,11 @@ mod tests {
         // install cannot re-read is declined, so the compiler's if/else-if chain
         // over the real parameter survives (option off = `switch(0)`, every case
         // unreachable and the parameter absent from the prototype)
-        assert_eq!(count, 238, "corpus file count drifted");
+        // and gh403-funcboundflow-cfg / the funcboundflow truncation of a
+        // CONDITIONAL branch's fall-through starts its own basic block, so the
+        // branch no longer self-edges (invalid `} while ;`, or a collapsed single
+        // out-edge the two-way block readers index off the end of)
+        assert_eq!(count, 239, "corpus file count drifted");
     }
 
     /// ~20 representative SLEIGH spec files across varied processors
--- a/decompiler/crates/kuna-decomp/src/p2_lift/flow.rs
+++ b/decompiler/crates/kuna-decomp/src/p2_lift/flow.rs
@@ -1470,7 +1470,13 @@ truncating the fall-through here"
                 // one (the cross-function merge bug).  Mirrors the noreturn-call
                 // halt in `check_for_flow_modification`; the halt is appended
                 // after the current instruction's ops by `artificial_halt`.
-                let _halt = self.artificial_halt(curaddr, pcodeop_flags::noreturn)?;
+                // It starts a basic block because `collect_edges` emits a
+                // fall-through edge for a CBRANCH regardless, and without a
+                // boundary that edge is a self-loop on the branch's own block.  It
+                // is NOT marked an instruction start: `fallthru_op` can only reach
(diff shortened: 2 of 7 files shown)


mahaloz:
The Tests workflow did not auto-trigger on this PR (no run was created for either `opened` or `reopened`), so it was started manually on the branch: https://github.com/Noelo-Lab/kuna/actions/runs/33979492429 — parity gates and the cargo workspace suite both green.

Issue #468, opened 2026-09-06
Friction report from a first kuna session (crackme + MSVC testbed, re-verified on v1.329)


CloudyTabzy:
Hi, just wanted to try Kuna for my real RE work today. Genuinely impressed and love what Kuna can do. Anyway, me and my agent are playing around with a `crackmes.exe` for a starter test and found some real friction worth reporting. Everything below was checked three ways: repro on v1.267, a git-log cross-check, and a re-verification on today's v1.329. (We initially split this into several issues to match the per-finding style the tracker usually sees, meant no harm by it — consolidating everything into this one issue instead.)

**Environment**: `kuna-v1.329-windows-x86_64` + `kuna-v1.329-specs`, Windows 11 x64, PowerShell 7.
**Testbed**: a purpose-built `ab_test.c` compiled twice with MSVC 18 (`/Od /GS /Zi` → `ab_od.exe`, `/O2 /GS /Zi` → `ab_o2.exe`, PDBs present) so every repro is byte-for-byte reproducible; the crackme and testbed binaries are available on request.
**For balance** — the same session also confirmed the good stuff: `iteboolean`, `branchflip`, `switchmultipred` (MSVC Duff's-device `memcpy` recovered as folded switches), region structuring + SAILR rendering a goto-ladder cleanup with zero gotos, and #409's `.pdata` work landing exactly where our v1.267 evidence pointed.

Status table:

| # | Finding | v1.329 |
|---|---|---|
| 1 | Windows zip cannot locate its own `decomp_dbg` (sibling probe misses `.exe`) | still broken |
| 2 | `loweredswitch` re-roll drops the switched value: `switch(0)`, parameter removed | still present |
| 3 | `stackguard` does not match the MSVC `/GS` canary shape (glibc-only matcher) | still present |
| 4 | `loweredswitch` misses the compiler-inlined cascade in an `/O2` caller | still present |
| 5 | `pdb` pass off by default, no sibling-`.pdb` discovery (the `.pdata` drift half was fixed by #409 — thank you) | still present |
| 6 | inlined `std::string` SSO appends render as indistinguishable duplicated blocks; `GS:[0x60]` PEB deref unnamed | still present |
| 7 | pointer-walk loop bound not derivable from the C render | still present |
| 8 | clean-run `int3` pad renders as a call (`swi(3)`) | still present (minor) |

## 1. Windows zip cannot locate its own `decomp_dbg` (sibling probe misses `.exe`)

[...]

PR #612, opened 2026-09-13
[AUTOMATED] fix(analysis): reject function starts interior to a PDB procedure (pdbinterior)

mahaloz:
## The problem

On the `/O2` testbed from #468 (`ab_testbed.zip`, `.pdb` beside the exe), `cascade_switch` is a leaf with no caller and no `.pdata` record, so only the PDB names it. The default mode starts two functions inside it and the decompiled switch loses its `0xf043` case.

```
$ kuna functions ab_o2.exe | awk -F'\t' '$1>="0x1400010e0" && $1<"0x1400011e0"'
0x1400010e0	cascade_switch
0x140001178	sub_140001178
0x1400011bc	sub_1400011bc
$ kuna decompile ab_o2.exe cascade_switch
  if (0x801d < a0) {
    if (0xc035 < a0) {
      if (a0 == 0xd03b)
        return 599;
      if (a0 == 0xe03d)
        return 0x263;
    }
    if (a0 == 0xc035) // warn: funcboundflow: fall-through reached the next function entry; truncating flow here
      return 0x215;
```

The PDB records `cascade_switch` as a 246-byte procedure (`[0x1400010e0, 0x1400011d6)`), so both entries are inside it.

## The fix

- New P1 pass `pdbinterior` (`on|off`, default on): reads each `S_GPROC32`/`S_LPROC32` code length from the fingerprint-matched `.pdb` and rejects discovered entries strictly inside a procedure, the same way `pdatainterior` uses `.pdata`. It only applies while `pdb` is on.
- A procedure is used only if no other procedure start, PDB public or image symbol lies inside it. A `.pdata` start inside a procedure does not disqualify it, because MSVC chains several unwind records within one function.
- `.pdata` and PDB bodies can nest, so the shared suppression now drops an entry that is inside any reported body instead of assuming the list is disjoint. The result is the same when the list is disjoint.

After the fix, `cascade_switch` has no `sub_` entries inside it and renders `if (a0 == 0xf043) return 0x2a5;`. Across `ab_o2.exe` / `ab_od.exe`, 101 / 99 bogus starts go away and none are added. 46 / 45 functions change, and each one had been truncated by `funcboundflow`: the warning count across the image drops from 46 to 0 and from 45 to 0. Every other function's C is byte-identical.

## The tests

- `pe_pdbinterior_x86_64.py` generates a 2 KB PE and a matching 7.7 KB `.pdb` that reproduce the defect: a PDB-only leaf whose fall-through `mov eax,10 ; mov edx,-1` matches the fingerprint of twenty helpers.
- `verify_pdbinterior.rs`: with the option off, `sub_1400011d0` is committed and the `0x1000` case is lost. The default removes the entry and renders every case. With `pdb off`, the extents are not applied.
- `tests/cli/pdb-leaf-interior-entry.json` runs `kuna decompile` on the fixture and fails against unpatched main.
- 675/675 datatest assertions unchanged. Speed: the module-stream walk costs +3.3 ms CPU (+1.9%) on `kuna functions --mode reliable` with the 6 MB PDB, and is within noise on the default path.

Part of #468.


mahaloz:
Adoption audit: the core PDB extent mechanism and committed fixture are credible, but this head is unsafe to land.

The blocker is local-procedure coverage: module walking uses `S_LPROC*` extents to suppress interior entries, while `PdbPass` currently admits procedure starts/names only from the global symbol stream. A false interior can therefore be removed without adding the PDB-known true local-procedure start/name, causing the only function representation to disappear. This head also has a stale DIV collision with current `main`, no trusted exact-head reviews or full Cargo result, and no current-base proof. It closes no registered RE need.

The source branch is intentionally preserved for a future Sol-high repair. That repair should either admit local PDB procedure starts/names before suppression or conservatively suppress only procedures whose starts are already admitted, with an `LPROC` fixture and an executable-range guard. It must then be freshly rebased, reallocated, fully validated, and independently reviewed.

PR #614, opened 2026-09-13
[AUTOMATED] feat(p6): endptrbound -- render a pointer walk's end bound on its own buffer

mahaloz:
A loop that walks a stack buffer with a pointer stops on the address one past the buffer, and kuna names that address after whichever local starts there, so the trip count (the buffer size) is not in the C. This is finding 7 of #468 (the crackme's 8-byte password loop), reduced to its bytes:

```bash
printf '555657488d6c24994881eca000000031c04889452f4889453f48c745470f000000c745273e2f3c3ec7452b3630382e488d75270fb616488d4d2fe81700000048ffc6488d452f4839c675e84881c4a00000005f5e5dc3488b411088140148ffc048894110c3' | xxd -r -p > walk.bin
kuna decompile walk.bin 0x140001000 --raw-image --target x86:LE:64:default:windows --base 0x140001000
```
```c
  unsigned int v1; // stack - 0x58
  unsigned long long v2 [2]; // stack - 0x50
  unsigned long long *v3;
  ...
  v1._0_4_ = 0x3e3c2f3e;
  v1._4_4_ = 0x2e383036;
  v3 = &v1;
  do {
    sub_140001056(v2,*(char *)v3);
    v3 = (unsigned long long *)((long long)v3 + 1);
  } while (v3 != v2);
```

`v2` is the object after the buffer, and the eight bytes became an `unsigned int` written through `_4_4_`. With this change the same function renders `char v1 [8];`, `char *v3;` and `} while (v3 != &v1[8]);`. Real MSVC code hits it too: `kuna decompile ab_o2.exe 0x1400373a4 --addr` (UCRT `write_string_to_console`) declares `char v4 [512]` where the PDB says `char narrow_buffer[500]`; it now renders `char v3 [500]`.

## Fix

- New option `endptrbound` (P6, default on, DIV-172).
- While the stack frame is laid out, recognize a pointer that starts at a stack address, steps by a constant, is dereferenced at that width, and is compared (`==`, `!=`, `p < end`) against a stack address a whole number of steps later. The frame hints inside that range become one array, unless a hint straddles an edge, is type-locked, holds a typed non-constant value, or carries index evidence past an edge.
- Once that array exists, the compared address is rebuilt as `&buf + span`. When the same address is also used for something else, such as the neighbour passed to a call, only the comparison is rebased, so that use keeps naming the neighbour.

## Tests

- `tests/stages/kuna-endptrbound.xml`: the shape above, a variant whose end register is also the call argument, and a control where a store straddles the end (identical in both passes). Pass 1 (`off`) pins today's render.
- Unit tests for the vetoes. 0/675 datatest assertions moved. `decompile-all` over 30 binaries (9,331 functions) changes 4 bodies outside the fixtures, all toward the source; CPU time is within 1%.

Part of #468.


mahaloz:
Sol-high adoption audit: closing without merge. The current recognizer does not prove that the candidate comparison controls the loop exit/backedge, so a midpoint or otherwise non-controlling comparison can invent an array extent and coalesce unrelated frame hints. It also records a phi as seen before confirming a matching access width, which can suppress a later valid access. The branch is stale: DIV-172 is already occupied by landed #611, and ElementId 4166 conflicts with pending work. It lacks direct detector/rewrite negatives, a promoted CLI acceptance, durable corpus/timing evidence, independent reviews, and a non-skipped full cargo run. This closes no registered repipe need. The source branch is intentionally preserved for a future Sol-high repair and current-main rebase, followed by fresh exact-head evidence and two Sol-low reviews.

Issue #182, opened 2026-07-28
Undersized open stack array: indexed store renders out of bounds; tail elements split off


mahaloz:
## Summary

For a stack array accessed through a computed index, kuna caps the recovered array at **4 elements** and splits every element beyond index 3 into separate scalar locals. The printed C is then self-contradictory in two ways at once. In betaflight `cliServo` it declares `int2 v5 [4];` and then emits `v5[v2] = sub_8017d38(a1);` under a guard `if (5 < v2) { ... return; }` — so the subscript provably reaches **5**, two past the declared bound. The two elements it runs past are declared as `int2 v6; // stack - 0x1c` and `int2 v7; // stack - 0x1a` — by kuna's own offset comments exactly `v5 + 4*2` and `v5 + 5*2` bytes, i.e. elements 4 and 5 of the same object — and **neither is ever assigned anywhere in the function**, yet both are read in the range guard and stored into the config struct. The binary has one 6-element `int16` array at `sp+28` written by one indexed `strh`; the source has `int16_t arguments[6]`. Ghidra prints a single `short local_24 [8]` and gets every index in bounds.

Root cause is a known seam, not a mis-port: `Funcdata::gather_open` supplies only the upstream fallback bound `min_items = 3`, and omits the `MapState::addGuard` LOAD/STORE-guard loops that would supply the real bound — because kuna's store guards are never range-refined (`analyzeNewLoadGuards` / `LoadGuard::establishRange` / `finalizeRange` / `ValueSetSolver::solve` are all documented stubs). `RangeHint::attempt_join` then rejects the `stack-0x1c` hint on `diffsz(4) > highind(3)` — off by exactly the missing bound.

## Reproduce

```bash
make binaries   # or: cd decompiler && cargo build --release

# Primary witness — ARM Cortex-M Thumb, gcc -O2, stripped
./decompiler/target/release/kuna decompile-all \
    tests/bug-repro/betaflight_STM32F405.elf --addr 0x801b4a4

# The same C is what the recompile-oriented project export writes:
./decompiler/target/release/kuna decompile-project \
    tests/bug-repro/betaflight_STM32F405.elf -o /tmp/proj --addr 0x801b4a4
grep -n 'v5 \[4\]\|v5\[v2\]\|v6\|v7' /tmp/proj/betaflight_STM32F405.elf.c

# Second witness — x86-64, gcc -O2, stripped (not staged in-repo)
./decompiler/target/release/kuna decompile-all \
[...]

PR #615, opened 2026-09-13
[AUTOMATED] fix(p6): keep a stack buffer's NUL terminator inside the array (nulterminator, DIV-173)

mahaloz:
## The problem

A stack buffer terminated with an explicit zero past its fourth element prints one element short, and the terminator write becomes a separate `char` that nothing reads. With the `ab_testbed.zip` from #468:

```
$ kuna decompile ab_od.exe canary_check
  unsigned char v3 [63];
  ...
  char v8; // stack - 0x19
  ...
  strncpy(v3,a0,0x3f);
  v7 = 0x3f;
  v8 = 0;
  for (v5 = 0; v3[v5]; v5 = v5 + 1) {
```

`kuna decompile ab_o2.exe main` has the same split (`unsigned char v10 [63]`, `char v15; // stack - 0x19`, `v15 = 0;`). The source is `char buf[64]; strncpy(buf, in, 63); buf[63] = 0;`.

## The fix

- New option `nulterminator` (P6 stack-frame layout, default on). When the layout sweep would end an open `char`/UTF-16 stack array at a one-element constant store, the array continues over it if every write to that slot is a zero and nothing reads the slot directly. Such a byte can only be read through the array pointer.
- The array's index bound is not raised, so a local that is read right after the terminator still starts a new variable, and wider elements are left alone (a zeroed `std::string` length field after its SSO buffer is not a terminator).
- The GH-182 fix (`loadguardrange`) cannot catch this: the terminator's constant index is folded into a direct stack store, and the scan loop's index is unbounded, so there is no guard range to lock.
- The `decompile-all --json` `variables` surface now credits an array with its element-store lines (`v2[8] = 0` left fauxware's `char v2 [12]` with no `line_numbers`).

After:

```
  unsigned char v3 [64]; // stack - 0x58
  ...
  strncpy(v3,a0,0x3f);
  v7 = 0x3f;
  v3[0x3f] = 0;
```

`--option nulterminator off` restores the old layout.

## Tests

- `tests/stages/gh468-nulterminator.xml`: the real `canary_check` bytes in two passes (off = `[63]` + split char, on = `[64]` + `v3[0x3f] = 0`), plus a hand-assembled neighbour whose zero byte is read back and keeps its own local in both passes. Unit tests cover the predicate.
- The 675 datatests render byte-identical C with the option on and off. Across 50 real binaries (19,222 functions) 6 functions change, all this idiom: both `canary_check`s, the `/O2` `main`, grep's two-char `strcoll` buffers, mirai `util_itoa`, cronie `load_database`.
- fauxware's `char stored_pw[9]; stored_pw[8] = 0;` now prints `char v2 [12]` instead of `char v2 [8]` + `char v3`: `realtypes-fauxware.xml` (a type-spelling test) pins the option off, and two console tests take the new layout.
- Speed: per-decompile CPU is unchanged within noise (`main` 25.1 ms both ways, `canary_check` 4.2 ms both ways); whole-binary `decompile-all` on sort/grep moves +0.4% / -1.3%.

Part of #468.


mahaloz:
[AUTOMATED] Sol-high ownership audit: closing without merge because the layout mechanism is unsafe.

The final PR binary turns Mirai `util_itoa` from its source-ground-truth `char scratch[34]` into `char[40]`, inventing six bytes. It also grows fauxware `stored_pw[9]` to `[12]`. The cause is structural: `absorbs_terminator` accepts an unread zero store at any later whole-element offset, while `ScopeLocal::restructure` skips that boundary without capping the current array extent. A later range hint can therefore absorb unrelated padding or a neighboring slot. The six advertised emitted-C changes do not establish that the zero belongs to the preceding array, so they cannot justify the transformation. DIV-173 is also already owned by merged PR #613.

The unrelated JSON provenance idea was independently reimplemented from current main, without any layout code, and will be proposed separately against registered need `address-taken-stack-buffers`. This PR remains available as evidence on its existing remote branch.

PR #622, opened 2026-09-13
[AUTOMATED] Honor explicit branch assertions over tail-call inference

mahaloz:
## Reproduction

An accepted explicit `flow 0x40d126 branch` was reinterpreted by `tailcalljump`, leaving only a tail-call stub.

## Fix

Track successfully applied BRANCH overrides by exact instruction address and resolve their precedence centrally before either `tailcalljump` or `tailcallframe` is consulted. Refused facts, other addresses, and other override kinds retain existing behavior. The exact PE32 witness is vendored and promoted as an executable CLI probe.

## Validation

Tail-call family tests pass 35/35, including a positive frame-teardown/no-assert control; CLI regression tests pass 3/3, the promoted acceptance passes 3/3, and `make check-spec` passes. Release A/B against parent `5fb5b031` is byte-identical for no assertion, `tailcalljump off`, and same-site `callreturn`; only the target BRANCH arm changes from the stub to the full body containing `0x40d12b` and `sub_40d21e()`.


mahaloz:
Superseded by a replacement draft PR to obtain a fresh GitHub CodeQL lifecycle. This PR passed exact-head parity, the non-skipped Cargo workspace suite, and two independent reviews; four CodeQL analyzers completed analysis/SARIF export but failed only at service-side upload, and GitHub rejected every rerun path.

PR #639, opened 2026-09-15
[AUTOMATED] fix(p3): name the real overlap when rename_recurse's input creation declines

sjkim1127:
## The problem

`rename_recurse`'s SSA input creation can decline and abort a function with a
fixed, content-free panic string, regardless of *why* it declined:

```
rename_recurse: set_input_varnode (empty stack)
```

That happens because `kuna_inputtile::new_tiled_input` discards the real
`LowlevelError` on the line that raises it:

```rust
// decompiler/crates/kuna-decomp/src/p3_dataflow/kuna_inputtile.rs:44 (before)
Err(_) => {
    let _ = fd.delete_varnode(candidate);
}
```

`docs/re-needs/ssa-renaming-panics-protected.md` tracks a live instance of
this (`thief_crackme.exe` @ `0x40ed14`), and its own decision log says the
next step is exactly this: "one eprintln at kuna_inputtile.rs:44 would settle
[which storage overlaps which input]". A cargo test against this crate shows
the same shape without needing that binary:

```
$ cargo test -p kuna-decomp p3_dataflow::kuna_inputtile::tests::new_tiled_input_declines_fully_contained_live_piece -- --nocapture
```
```
Overlapping input varnodes (requested input ram[0x100,0x104);
existing input [0x101,0x102) is fully contained in the request but is a live
(non-write-masked) input, not a guardInput leftover piece)
```
vs. the old `rename_recurse: set_input_varnode (empty stack)` a caller would
have seen with no other detail.

## The fix

- `new_tiled_input` / `combine_covered_pieces` return `KunaResult<VarnodeId>`
  / `Result<VarnodeId, String>` instead of `Option<VarnodeId>`, so the
  original `LowlevelError` and a specific reason (straddling vs.
  fully-contained, write-masked vs. live, or the requested range itself)
  reach the caller instead of being dropped.
- The three `rename_recurse` call sites fold that into their panic message
  instead of a bare `.expect("...")`.
- Pure diagnostics: the reconciliation acceptance criteria and successful
  transformation paths are unchanged, so every existing success path is
  unaffected.

This PR intentionally does not resolve `ssa-renaming-panics-protected`; it
makes the currently unknown overlap geometry observable so the eventual
reconciliation fix can be made safely, against a real witness and a
whole-corpus sweep.

## The tests

Two new unit tests against `kuna_inputtile` directly (no CFG/heritage-round
scaffolding needed, following the existing `build_fd`/`make_input` pattern
from `funcdata_varnode.rs`'s own tests):
- a live, non-write-masked, fully-contained piece is still declined, and the
  message names both the requested and conflicting ranges;
- the DIV-50 write-masked-piece reconciliation this module exists for is
  regression-pinned (still combines correctly).

`make test` (675/675), `make test-stages` (1003/1003), `make check-spec`, and
`make rust-test` all pass unchanged.


mahaloz:
[AUTOMATED] Merged. Reviewed the Option→Result conversion path by path: no accept/decline decision moved, no success path allocates (every message is built inside an `ok_or_else`/`map_err` closure), and the three `rename_recurse` sites panic in exactly the same states as before. The new message is a superset of the old string, so the `ssa-renaming-panics-protected` reproduction probe still matches.

Two nits left as-is, neither worth a round trip: `kuna_inputtile.rs:58` builds the outer message's `hi` with `wrapping_add` while the body uses `checked_add`, so an overflowing request prints a wrapped range next to its "overflows" reason; and the doc comment hard-codes "the three `rename_recurse` call sites currently turn this into a bare panic", which rots when the need is actually closed.

PR #643, opened 2026-09-16
[AUTOMATED] fix(p9): keep a float expression's grouping instead of flattening it

### LINKED ISSUE #641: Floating-point multiplication and addition lose grouping (associativity) in emitted C
Found an issue with Astra where associativity of operations in floating point impacted the result even under `mode reliable`. Essentially it's the old issue where binary operations in fp arithmetic are not associative but the decompiler assumes they are. Attached is a bundle for reproducing the issue too, the rest of the report is what I asked Astra to make. If needed I can point it to write a PR, in any case feel free to ping me.

## Astra's report
Two x86-64 functions with different IEEE-754 rounding order decompile to
identical C in reliable mode. Native execution produces different float bits.
The input is a small, independently authored assembly fixture; no game
executable or third-party assets are needed.

Tested with Kuna source revision
`fa88f98fe8ca839408c73922646c613a6d41dccf`, Linux x86-64, C output,
`--mode reliable`, no assertions. Kuna binary SHA-256:
`68c429888b7047652dca027b23f095caac1f3f6f16c96c361a6dd00a0de47988`.

## Reproduce

Save as `repro.S`:

```asm
.text
.global right_grouped
.type right_grouped, @function
right_grouped:
    mulss scale(%rip), %xmm1
    mulss %xmm1, %xmm0
    ret
.size right_grouped, .-right_grouped

.global left_grouped
.type left_grouped, @function
left_grouped:
    mulss %xmm1, %xmm0
    mulss scale(%rip), %xmm0
    ret
.size left_grouped, .-left_grouped

.section .rodata
.align 4
scale:
    .long 0x3fb33333
.section .note.GNU-stack,"",@progbits
```

```sh
cc -shared -nostdlib -o repro.so repro.S
nm -D repro.so
kuna decompile-all repro.so --addr 0x1000 --addr 0x100d --mode reliable --json
```

`nm` gives `right_grouped=0x1000`, `left_grouped=0x100d` with the tested
linker; use its actual addresses if your linker places them differently.
System V arguments are `factor` in XMM0 and `speed` in XMM1; the constant is1.4f.

Actual output:

```c
float4 right_grouped(float4 a0,float4 a1)
{
  return a0 * a1 * dat_2000;
}
float4 left_grouped(float4 a0,float4 a1)
{
  return a0 * a1 * dat_2000;
}
```

Expected: the first return should retain `a0 * (a1 * dat_2000)` or an
explicit float temporary. The second return can remain left-associated.

## Observable difference

Run against the same shared library:

```python
import ctypes, struct
lib = ctypes.CDLL('./repro.so')
for name in ('right_grouped', 'left_grouped'):
    fn = getattr(lib, name)
    fn.argtypes = [ctypes.c_float, ctypes.c_float]
    fn.restype = ctypes.c_float
    value = fn(1.0185916423797607, 5.77)
    print(name, value, '0x' + struct.pack('>f', value).hex())
```

```text
right_grouped 8.22818374633789 0x4103a6a4
left_grouped  8.228182792663574 0x4103a6a3
```

Both inputs and results are finite normal floats. This needs no NaNs,
signed-zero edge cases, overflow, FMA or fast-math compilation. The functions
are assembled directly, preserving both MULSS rounding points.

## Likely printer cause

At the tested revision, `CPUI_FLOAT_MULT` uses the same `MULTIPLY` token as
integer multiplication; that token is marked associative. The parentheses
routine omits parentheses when nested tokens match and are associative.
This is a source-level lead consistent with the fixture, not a verified patch:

- [Float opcode and operator definitions](https://github.com/Noelo-Lab/kuna/blob/fa88f98fe8ca839408c73922646c613a6d41dccf/decompiler/crates/kuna-decomp/src/p9_emit/printc.rs)
- [Parentheses logic](https://github.com/Noelo-Lab/kuna/blob/fa88f98fe8ca839408c73922646c613a6d41dccf/decompiler/crates/kuna-decomp/src/p9_emit/printlanguage.rs)

Current upstream main `39837a2c17fcfca458acb4a5015fe282bb1ffaaf`, inspected
on2026-09-15, still contains these same token mappings and parentheses check.
That is source inspection only; the executed binary is the pinned revision above.

## Addition also reproduces

The attached `repro_add.S` uses three float arguments in XMM0/XMM1/XMM2:

```asm
right_add:
    addss %xmm2, %xmm1
    addss %xmm1, %xmm0
    ret
left_add:
    addss %xmm1, %xmm0
    addss %xmm2, %xmm0
    ret
```

Build using `cc -shared -nostdlib -o repro_add.so repro_add.S`; the tested
addresses are0x1000 and0x1009. Reliable-mode output prints both returns as
`a0 + a1 + a2`. Native execution with float inputs `(1e20, -1e20, 3.14)` gives
0.0 (`0x00000000`) for right_add and3.140000104904175 (`0x4048f5c3`) for left_add.
`CPUI_FLOAT_ADD` likewise uses an associative operator token. A prospective
regression should distinguish left/right float expression trees without
unnecessarily changing integer formatting.

## Existing issue search

Checked GitHub on2026-09-15: no matching open issue found among the three open
issues, or in issue/PR searches for float, floating, mulss, reassociation,
rounding, parentheses and precision. [#275](https://github.com/Noelo-Lab/kuna/issues/275)
is a closed floating-point argument-recovery issue, not this grouping problem.

This file is a local draft; it has not been submitted upstream.

[upstream_bundle.zip](https://github.com/user-attachments/files/32261093/upstream_bundle.zip)

mahaloz:
## The problem

Two functions that multiply the same three floats in different orders decompile
to the same C. Assemble the pair and decompile both:

```sh
cat > repro.S <<'ASM'
.text
.globl right_grouped
.type right_grouped, @function
right_grouped:  mulss %xmm2, %xmm1; mulss %xmm1, %xmm0; ret
.globl left_grouped
.type left_grouped, @function
left_grouped:   mulss %xmm1, %xmm0; mulss %xmm2, %xmm0; ret
.section .note.GNU-stack,"",@progbits
ASM
cc -shared -nostdlib -o repro.so repro.S
kuna decompile repro.so right_grouped
kuna decompile repro.so left_grouped
```

```c
float right_grouped(float a0,float a1,float a2)
{
  return a0 * a1 * a2;
}
float left_grouped(float a0,float a1,float a2)
{
  return a0 * a1 * a2;
}
```

`right_grouped` computes `a0 * (a1 * a2)`, which rounds at different points.
The reporter's pair returns `0x4103a6a4` and `0x4103a6a3` from the same
arguments, so the first of those two bodies is C that computes a different
value than the function it came from. `addss` reproduces it the same way, and
so does `--mode reliable`.

## The fix

- `CPUI_FLOAT_MULT` and `CPUI_FLOAT_ADD` push their own token singletons,
  spelled and ranked exactly like `MULTIPLY` and `BINARY_PLUS` but carrying a
  new `left_to_right_only` instead of `associative`.
- That flag is a third answer `parentheses()` can give a same-token operand:
  `associative` leaves both bare, a plain non-associative token (`-`, `/`)
  parenthesises both, and this one leaves the left bare and parenthesises the
  right. It is the exact rule rather than the conservative one, because C
  already groups `*` and `+` left-to-right, so only a right operand needs the
  parentheses to survive the round trip — `left_grouped` above is unchanged.
- Float subtraction and division already reached the right answer through the
  non-associative tokens they share with the integer operators.

## The tests

`tests/stages/kuna-floatgrouping.xml` decompiles both groupings of both
operators: two assertions pin the parenthesised form, two pin that the flat
chain now belongs to exactly one of each pair, where before this change it
matched both. Four unit tests cover the token table, `parentheses()` directly,
and the RPN engine end to end.

0 of 675 datatest and 0 of 1003 stage assertions move.

Fixes #641

https://claude.ai/code/session_01PFgY2o2YeV9bjSbJLcdCs2


mahaloz:
[AUTOMATED] Merged.

An adversarial review swept this over all 879 functions of the system `libm.so.6`, A/B against main: 76 bodies changed, every change is purely parenthesisation (stripping all parens makes both identical), 76 gained parens and 0 lost any. `__jnl_finite` is a textbook instance — `v20 = v20 * v11 * (float10)dat_8a1cc;` becomes `v20 = v20 * (v11 * (float10)dat_8a1cc);`. Reading `parentheses` confirms why that is guaranteed: the new `return false` only fires where main already returned false, so the identity split can only ADD parentheses, never remove one, and therefore cannot turn correct output into incorrect output.

Worth recording about the "0 of 675 datatest assertions move" evidence: `tests/datatests` does exercise float `+`/`*`, but only as single operators (`fval + 5.5`, `myfloat * 1.5`) — there is no nested same-token float chain in that corpus, so those assertions could not have moved whatever the change did. The real non-regression signal is `tests/stages/kuna-msvcfpconst.xml`, which carries a three-operand chain `(float10)dat_402010 * (float10)a0 * (float10)dat_400040` and correctly stayed flat because it is left-nested.

One direction where output changes without a bug being fixed: an `INT_ADD` nested directly under a `FLOAT_ADD` with no intervening `CPUI_CAST` now gains parentheses it does not need. Both forms are correct C, and it appears in none of the ~1,850 gate assertions nor the libm sweep — casts normally sit between the two type worlds.

PR #644, opened 2026-09-16
[AUTOMATED] fix(cli): decompile a streamed export's seeds in a worker, not the parent

### LINKED ISSUE #636: [AUTOMATED] decompile-project --stream aborts the whole export when a seed function overflows the stack
`decompile-project --stream` decompiles its seed functions (the entry and `main`) in the parent process before the worker pool starts. A seed that overflows the stack therefore aborts the whole export, where `--jobs` would lose one record. crackmes.one `684a1a2a2b84be7ea7743801` (`AmeenOggly.exe`):

```
$ kuna decompile-project AmeenOggly.exe --stream --jobs 4
[kuna --jobs] decode: serial (executable image too small)

thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting
$ echo $?
134
```

The export is left with an empty `.asm` and `index.jsonl`, a 29-byte `.c`, and a `.streaming` marker that still says `"phase":"decompiling"` with a dead pid (0 of 1,031 functions done). `decompile-all AmeenOggly.exe --jobs 4` on the same binary finishes, with the overflowing function reported as a lost record (and since #631 its chunk-mates are re-run).

A stack overflow cannot be caught in-process, so the likely fix is to decompile the seeds in a worker as well, or to fall back to a worker when the in-process seed decompile is not possible.

https://claude.ai/code/session_01Ec4F113Qm7hnvUycXdFLbL

mahaloz:
## The problem

`decompile-project --stream` decompiled the entry point and `main` in the
parent process, before the worker pool started. A seed that took its process
down took the whole export with it. On crackmes.one `684a1a2a2b84be7ea7743801`
(`AmeenOggly.exe`):

```
$ kuna decompile-project AmeenOggly.exe --stream --jobs 4
[kuna --jobs] decode: serial (executable image too small)

thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting
$ echo $?
134
```

The export folder is left with an empty `.asm` and `index.jsonl`, a 29-byte
`.c`, and a `.streaming` marker still reading `"phase":"decompiling"` with a
dead pid, 0 of 1,031 functions done. `decompile-all` on the same binary
finishes, because the function that overflows runs in a worker there.

## The fix

- `run_pooled` no longer decompiles the seeds itself; they go to the pool with
  every other function. A stack overflow aborts rather than unwinding, so a
  worker is the only place that can contain one.
- That is paid for in the order. The seeds are still the head of the frontier,
  so they still go out first, but nothing fills that frontier before the pool
  starts — so at `--jobs N` the workers that do not draw the seeds spend their
  first round on the address cursor, and the seeds' neighbourhood lands later in
  the `.c`. On the `fauxware` fixture at `--jobs 2`, `main`'s direct callees move
  from the fourth block to the last. `--jobs 1` keeps the breadth-first walk
  exactly, and is also the one job count this does not protect: with no pool the
  parent decompiles everything itself. Both are in `docs/cli.md` and the spec,
  and #645 tracks closing the gap.
- `done_base` and `Scheduler::claim` went with the in-process decompile; both
  callers passed `done_base: 0`.

I tried making the address cursor wait for the seeds' results instead of
accepting that, and removed it: measured, the worker that served the seeds is
warm and wins the first frontier chunk, then takes the 64-wide cursor while the
woken worker is still loading — so `main`'s callees still landed last — and the
wait's 30 s floor sits under the 120 s per-function budget, so any binary that
loads in under 40 s could stall the whole export for 30 s with nothing written.
Details in #645.

The reported binary now finishes: `1031/1031 functions ... 1019 ok, 12 failed`,
with the overflowing function one `worker chunk failed (worker exited: signal:
6 (SIGABRT))` record, its chunk-mates re-run and recovered, and no `.streaming`
left behind.

## The tests

`a_streamed_seed_runs_in_a_worker_and_costs_only_its_own_record` arms
`KUNA_JOBS_FAULT=panic:0x400580` — fauxware's entry point, its first seed — and
asserts the export completes, the seed's record carries the worker failure, and
no bystander's body moves. The fault hook lives in the worker, so before this
change the panic fired nowhere at all and the export came back clean; the test
fails there on exactly that.

Fixes #636

https://claude.ai/code/session_01PFgY2o2YeV9bjSbJLcdCs2


mahaloz:
[AUTOMATED] Merged.

Two reviews shaped this. The first pointed out that handing the seeds to the pool also removed the only thing that filled the frontier before the pool opened, so every other puller took up to 64 address-ordered targets before the seed neighbourhood existed — on `fauxware` at `--jobs 2`, `main`'s direct callees moved from the fourth block of the `.c` to the last.

I tried fixing that by making the address cursor wait for the seeds' results, and the second review killed it, correctly:

- the deadline was computed on entry to the wait, so it re-armed on every empty-frontier pull — a repeated 30 s stall, not a one-time round. Measured with a stalled seed: two full 30 s gaps with zero output, 69 s for an export that takes 2.7 s clean.
- "the deadline only fires if a chunk's results never arrive at all" was false. `opening_wait` floors at 30 s while the default per-function budget is 120 s, so any binary loading in under 40 s could hit it on a perfectly healthy run — a total output stall on the one flag whose point is early output.
- and it did not even work: the worker that served the seeds is warm, wins the first frontier chunk, then takes the 64-wide cursor while the woken worker is still spawning and loading, so `main`'s callees still landed last at `--jobs 2`.

So this merges as the crash fix alone, with the ordering cost stated plainly in `docs/cli.md` and `docs/spec/00-overview.md` rather than papered over. #645 has the measurements and what an actual fix would need (overlapping the pool's worker loads with the seeds' round, which is a change to the pool shared with `--jobs` and wants its own PR).

Verified on the reported binary: `1031/1031 functions ... 1019 ok, 12 failed`, the overflowing function one `worker exited: signal: 6 (SIGABRT)` record, its chunk-mates recovered, no `.streaming` left behind.

PR #670, opened 2026-09-17
[AUTOMATED] feat(p6): wire ActionHideShadow so redundant copies of one value collapse (hideshadow)

mahaloz:
## The problem

kuna prints the same assignment twice: a value copied into one variable on two
paths that both reach the use stays two separate writes, so the listing repeats
a statement the binary does not execute twice.

```console
$ cat > t.c <<'EOF'
void f(char *dest, const char *dirname, const char *name)
{
  const char *p = dirname;
  if (dirname[0] != '.' || dirname[1] != 0) { while (*p) *dest++ = *p++; }
  while (*name) *dest++ = *name++;
  *dest = 0;
}
void _start(void){ }
EOF
$ gcc -O0 -nostdlib -static -fno-pie -no-pie -o t t.c
$ kuna decompile-all ./t --addr 0x401000
void f(char *a0,char *a1,char *a2)
{
  char *v1;
  char *v2;
  char *v3; // stack - 0x30
  char *v4; // stack - 0x20
  char *v5; // stack - 0x10

  v4 = a0;
  v5 = a1;
  if ((*a1 != '.') || (v3 = a2, v4 = a0, v4 = a0, v5 = a1, a1[1])) {
```

`v4 = a0` twice, with nothing in between. Upstream Ghidra has a pass for
exactly this (`ActionHideShadow`, `coreaction.cc:4976`) and kuna ported its
engine half (`Merge::hideShadows`), but the action body was a stub returning 0,
so the pass never ran.

## The fix

- `p6_variables/kuna_hideshadow.rs` is the upstream body: walk the written
  def-set, visit each `HighVariable` once (deduped through the bank's mark
  bit), call `Merge::hide_shadows`. The action is already in `universal_sched`
  at its upstream slot, so no action index moves and no oracle is re-recorded.
- Both acceptance tests are upstream's: the two Varnodes must trace to one
  ancestor through COPYs alone, and the surviving one's Cover must contain the
  other's definition point *strictly interior* — so the value is live and
  unclobbered where the copy is made to read it. Copies on exclusive branches
  fail that and are left alone (over 40 instrumented binaries the Cover test
  accepts 110 candidates and declines 9,138).
- The Cover read is preceded by a `bank_update_cover` refresh. C++
  `Varnode::getCover()` rebuilds a dirtied cover on the spot, kuna's accessor is
  a plain read, and the previous iteration's `opSetInput` is what dirties it.
- Ships on, since it is what stock Ghidra does; `--option hideshadow off`
  returns before the walk, byte-for-byte what the tree did with the stub.

## The tests

`tests/stages/kuna-hideshadow.xml` is two-pass over the binary → decompile path
(off = the repeated copy, on = one copy) plus a third function whose two copies
sit on exclusive branches and must survive on. Whole-corpus `decompile-all`
A/B over 444 stripped ELFs: 434 byte-identical, 10 changed, and all 15 removed
assignments (12 deleted lines plus 3 lines rewritten to drop a duplicated term
from a comma expression) are ones an identical assignment already dominates, no
declaration moved (`docs/features/hideshadow/`). 0/675 datatest assertions move
with the flip, stages PARITY OK, `type_match` unmoved over 303 scored slices,
and decompile time is inside the noise: interleaved off/on/off, min of 15, the
witness function +1.66% and a 6.5 s function +0.62%, against off-vs-off controls
of -0.20% and -1.47%.

https://claude.ai/code/session_01MAeimyWm5oSewpeCgAwY29


mahaloz:
[AUTOMATED] Review round 1 addressed.

**Counter collision (blocker).** Rebased three times as main moved — #661 structdefs, #663 indirectonly, #660 libctypes glibc — and re-derived the counter sites with `counters --rederive/--fix` on a freshly built tree each time rather than letting the identical stale numbers 3-way-merge: settables 216, tiers 68/85/63, catalog json rows 215, bytecompat fixture 216, stage corpus 306, live_value 88. `phase_catalog.json` recaptured, `docs/options.md` byte-fresh, `docs/baseline-stages.json` re-recorded (1078 -> 1081, the three hideshadow assertions and nothing else). #663 also touched the spec's phalanx-tail sentence; merged, so it now lists only `ActionMergeMultiEntry` as inert.

**`addresses` claim.** Confirmed and corrected — openssh sshd O2, 14,006 variables compared: 2 differ (fn 0x17630 v30 loses 0x1a106; fn 0x46cb0 v4 loses 0x46d31 and 0x46e87). `record.json` now says the hidden copy's address leaves the list and that no scorer reads the field.

**Line count.** `corpus-diffs.txt` is 15 `-` lines and 3 `+` lines: 12 statements deleted outright and 3 lines rewritten to drop a duplicated comma-expression term, i.e. 15 removed assignments, net -12 lines. Fixed in the PR body, `corpus-classification.md`, `analysis.md` and `record.json`.

**Doc drift and anchors.** Both arch doc comments now say default-on and match `reset_defaults`; the count-test comment says default-on; the section banner and struct doc cite `coreaction.hh:1004` / `coreaction.cc:4976` and describe this pass rather than CONCAT stacks.

**Speed.** Re-measured interleaved off/on/off in one loop, min of 15 per arm (`docs/features/hideshadow/speed-remeasure.txt`): ls O2 sub_7ff0 +1.66% against a -0.20% control, sshd O2 sub_17630 (6.5 s) +0.62% against a -1.47% control. The +3.68% reading does not reproduce.

Gates on the rebased head: `make test` 675/675 PARITY OK, `make test-stages` 1081/1081 PARITY OK, `make test-cli` 183/183, `check-spec` lenient + strict, `kuna catalog --check` OK, `counters --check` no drift, options.md byte-fresh, `make rust-test` RUST_TEST_RC=0 (416 `test result: ok`, 0 failures). CI green including the workspace suite. Also re-checked after the rebase: the OFF arm is byte-identical to a main build, all 10 known-changed slices still change with 38 random others identical, and with `indirectonly` on the diff is the same two hunks.

PR #679, opened 2026-09-18
[AUTOMATED] feat(p5): structsynth keeps a homogeneous integer pair as an element pointer

mahaloz:
## The problem

With `--option structsynth param`, a parameter that is already a correct `long *` and is read at `[0]` and `[1]` becomes a structure of two `long`s. It is the same bytes under a record name, and it replaces a type that was right. Here are the eleven bytes of `p[0] += p[1]`:

```
$ printf '\x48\x8b\x07\x48\x03\x47\x08\x48\x89\x07\xc3' > limbs.bin
$ kuna decompile limbs.bin 0x100000 --raw-image --target x86:LE:64:default:gcc --base 0x100000 \
    --option structsynth param --option structdefs on
struct struct_0 {
    long field_0x0;
    long field_0x8;
};

void sub_100000(struct_0 *a0)
{
  a0->field_0x0 = a0->field_0x0 + a0->field_0x8;
}
```

Without the option it prints `void sub_100000(long *a0)` and `*a0 = *a0 + a0[1];`. Real binaries have the same shape. coreutils `factor`'s `powm2(uintmax_t *, const uintmax_t *, ...)` (decbench O2, `--addr 0x3190`) prints `struct_0 *a1, struct_1 *a2, struct_0 *a3, ..., struct_0 *a5` where every one of them is a `uintmax_t *`.

## The fix

- `structsynth` now declines a same-typed integer pair when it says nothing the pointer does not. All three conditions must hold. Every raw access is one width `w` at exactly `0, w, ...`. Every field would get the same integer type. The base already points at an integer of width `w`.
- It still synthesizes a pair of pointers (in the census, every such pair that DWARF can type is a record). It also still synthesizes a pair that differs in type or signedness, a pair with any access the layout prune drops, and a pair read through a pointer to some other element (`char *`).
- The default is unchanged (`off`). The default arm is byte-identical to `main` on 16 binaries, text and `--json`.

## The tests

- `tests/stages/structsynth-element-pair.xml`, two passes. `limbs` keeps `int8 *` with the option off and on, and a pointer-plus-integer `record` is still `struct_0`. On `main`, asserts 1-4 fail.
- Four unit tests in `kuna_structsynth/tests.rs`: signedness, pointer and float pairs, and a pruned, misaligned or holed access are all kept.
- 444-slice `type_match`, param vs off: on `main`, perfect goes 959 -> 958 and 7 rows get worse, 6 of them in `factor`. On this branch, perfect stays 959 -> 959 and 1 row gets worse (`shred` `passname`, a mixed-width `char[7]`, not a pair).
- The cost: layout F1 (fmt/ls/sort/du) goes 0.1376 -> 0.1340, and TRex Fig. 6 goes 2.6604 -> 2.6597. That is 20 fields that were right (`stat`'s `st_dev`/`st_ino`, `timespec`, `hash_entry`). If `type_match` credited any `struct_N *` against a DWARF structure pointer, the rule would lose 109 scored variables to gain 18 (net -91; `arrays_credit.txt`).
- Speed (`decompile-all`, interleaved min-of-15, param on main vs param here): +3.27% to -5.47% on factor/fmt/ls/du. That is inside the +/-5.4% an inert control moves between the same two builds.

Default-on numbers for a later flip (not flipped here): 675/675 datatests, 2 stage assertions move (PEBNAMES-X86 #6/#8, the intended TEB fields), `type_match` 959 -> 959, no dangling-`goto` hunk, speed at most +0.32% after a re-run. Details are in `docs/features/structsynth/arrays.md`.

https://claude.ai/code/session_01YFCA715QE6ZBSnTn7bwabG


mahaloz:
[AUTOMATED] Closing without merging. With structsynth on by default, and with struct pointers credited by kind rather than by name (Noelo-Lab/decbench#93), this decline removes more correct struct pointers than it fixes. It drops 196 struct_N pointers: 109 are real struct pointers in DWARF (hash_entry, stat, timespec, ...) and only 18 are the factor GMP limb arrays it targets. It also declines 4-byte pairs on 32-bit targets, and those were never measured. The factor false positive needs a narrower signal, such as a variable index into the base, not a width-uniform pair.

